PR #13, opened 2026-02-16
Support exporting from SQLite and HyperDB sites

adamziel:
## Summary

The export system has been hardcoded to MySQL — three call sites created `new PDO("mysql:...")` directly, and `resolve_db_credentials()` required MySQL-specific constants. SQLite-powered WordPress sites couldn't export at all.

This introduces a `create_db_connection()` factory that detects the database backend and returns the right connection. For SQLite, queries go through the `sqlite-database-integration` plugin's `WP_SQLite_Driver` wrapped in a thin PDO adapter. For HyperDB, the wp-config.php credentials already point to the write master, so the MySQL path works as-is. The output format stays the same — always valid MySQL SQL.

**Detection** happens in `api.php` by reading `wp-content/db.php` (contains "sqlite" → SQLite, "hyperdb" → HyperDB) and wp-config.php constants (`DB_ENGINE`, `DB_DIR`, `DB_FILE`).

**New files:**
- `class-sqlite-driver-pdo.php` — `SqliteDriverPDO` + `SqliteDriverPDOStatement`, a PDO-compatible adapter wrapping `WP_SQLite_Driver`
- `sqlite-bootstrap.php` — locates the plugin, includes its class files, polyfills `do_action`/`apply_filters`, wires up the driver

**Key changes to existing files:**
- `export.php` — new `create_db_connection()` factory replaces three inline PDO constructions; `resolve_db_credentials()` relaxed for SQLite (no host/user/password needed); preflight checks `pdo_sqlite` extension for SQLite sites and reports `db_engine`; MySQL `@@variable` queries wrapped in their own try/catch for graceful degradation
- `class-mysql-dump-producer.php` — PDO type hint removed from constructor to accept the duck-typed adapter
- `api.php` — extracts `DB_ENGINE`/`DB_DIR`/`DB_FILE` from wp-config.php, inspects `db.php` drop-in for backend detection

## Test plan

- [x] All 254 existing tests pass (`composer test`)
- [x] PHPStan analysis clean (`composer analyze`)
- [ ] Deploy to a SQLite WordPress site, trigger export, verify dump contains valid MySQL `CREATE TABLE` + `INSERT` statements
- [ ] Import the SQLite-exported dump into MySQL and compare row counts
- [ ] Verify standard MySQL export still works unchanged

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -9,6 +9,8 @@ parameters:
         - importer
     excludePaths:
         - wordpress-plugin/generic/secrets.php
+        - wordpress-plugin/generic/class-sqlite-driver-pdo.php
+        - wordpress-plugin/generic/sqlite-bootstrap.php
         - wordpress-plugin/wordpress/
         - wordpress-plugin/index.php
     treatPhpDocTypesAsCertain: false
--- a/wordpress-plugin/api.php
+++ b/wordpress-plugin/api.php
@@ -213,8 +213,13 @@ if ($wp_root !== null && !defined('DB_HOST')) {
     if (is_readable($wp_config_path)) {
         $config_contents = @file_get_contents($wp_config_path);
         if ($config_contents !== false) {
-            // Extract define('CONSTANT', 'value') patterns
-            foreach (['DB_HOST', 'DB_NAME', 'DB_USER', 'DB_PASSWORD'] as $const) {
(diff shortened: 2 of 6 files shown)


adamziel:
We'd need to migrate from using a separate PDO connection to using $wpdb. It can be done, but I'd rather avoid `$wpdb` as long as we can. I don't trust it and I'm not sure if it actually supports streaming the result set (even if it claims it does).

PR #16, opened 2026-02-17
Add URL rewriting for imported SQL data

adamziel:
## Summary

- Adds a `db-apply` command that reads `db.sql`, rewrites URLs using [wp-php-toolkit](https://github.com/WordPress/php-toolkit) structured processors (`BlockMarkupUrlProcessor`, `URLInTextProcessor`), and executes statements against a target MySQL database
- Domain discovery runs inline during `db-sync`, collecting all HTTP/HTTPS domains from decoded base64 values into `.import-domains.json`
- Uses `WP_MySQL_Naive_Query_Stream` (vendored from [WordPress/sqlite-database-integration#264](https://github.com/WordPress/sqlite-database-integration/pull/264)) for streaming SQL parsing with cursor-based resumability

### URL rewriting pipeline

For each base64-decoded string value in INSERT statements:
1. **Serialized PHP** → detected via `ContentClassifier` (port of `is_serialized()`) → **skipped** (adjusting `s:N:` length prefixes is out of scope)
2. **JSON** → decoded, all string values walked recursively, URLs rewritten via `URLInTextProcessor`, re-encoded
3. **Everything else** (HTML, block markup, plain text, markdown) → `wp_rewrite_urls()` which handles HTML attributes, block comment JSON, text nodes, and CSS `url()` in style attributes

No `preg_match` or `DOMDocument` — only the structured data processors from `wp-php-toolkit/data-liberation`.

### New components

| File | Purpose |
|------|---------|
| `importer/lib/Base64ValueScanner.php` | Scans SQL for `FROM_BASE64('...')` expressions, decodes values |
| `importer/lib/ContentClassifier.php` | Classifies values as serialized PHP, JSON, or text/HTML |
| `importer/lib/DomainCollector.php` | Collects unique domains from string values via `URLInTextProcessor` |
| `importer/lib/SqlValueUrlRewriter.php` | Rewrites URLs in a single value using the appropriate processor |
| `importer/lib/SqlStatementRewriter.php` | Orchestrates per-statement rewriting |
| `importer/lib/mysql-query-stream/` | Vendored `WP_MySQL_Naive_Query_Stream` + `WP_MySQL_Lexer` |

### Usage

```bash
# 1. Download SQL (domains discovered automatically)
php import.php db-sync https://source.example.com /path/to/import --secret=TOKEN

# 2. Apply with URL rewriting
php import.php db-apply - /path/to/import \
  --target-user=root --target-db=wp_new \
  --url-mapping=https://source.example.com::https://target.example.com
```

## Test plan

- [x] 72 unit tests across 5 test files (Base64ValueScanner, ContentClassifier, DomainCollector, SqlValueUrlRewriter, SqlStatementRewriter)
- [ ] E2E round-trip test (`import-36-url-rewriting.test.js` — scaffolded, needs implementation)
- [ ] Manual test with a real WordPress SQL dump


adamziel:
It seems like the domain detection could break if the download stream pauses mid-query, e.g.

```
INSERT INTO ... VALUES
(FROM_BASE64("STRING1")),
<paused right here>
```

So we'll need some way of resuming from there – even if that way is oversimplified for now and requires going back to the initial `INSERT`. Maybe that could happen in a follow-up PR, it's a relatively low stakes error. Worst-case scenario we'll report an incomplete list of detected domains. We'll still be able to replace any listed domains when inserting the data.

It would be nice to auto-detect the source site domain and mark it for rewriting, maybe assets domain as well if it's separate.

Other than that, I'll need to take this for a spin.

adamziel:
Also – it would be cool to support downloading the media files loaded from external domains, but we don't strictly need it to land this feature.

importer/import.php
```diff
@@ -4066,11 +4497,63 @@ private function download_sql(): void
...
+    private function drain_query_stream_for_domains(
+        \WP_MySQL_Naive_Query_Stream $query_stream,
+        \DomainCollector $domain_collector
+    ) {
+        while ($query_stream->next_query()) {
+            $query = $query_stream->get_query();
+            // Only scan INSERT statements (they contain data values).
+            // The query stream may prepend SQL comments to the statement,
+            // so check for INSERT anywhere rather than only at position 0.
+            if (stripos($query, 'INSERT INTO') === false) {
```
adamziel: so  naive, nope, we need something more excact than this

PR #26, opened 2026-03-03
Add --on-docroot-nonempty to import into non-empty directories

adamziel:
## Summary

Adds a `--on-docroot-nonempty=preserve-local` flag that allows importing into a non-empty directory while keeping all existing local content intact.

## Rationale

files-sync currently refuses to start if the target docroot is non-empty. This blocks a real use case: importing into a WordPress hosting environment where plugins, themes, drop-ins, and WP core are symlinked from a shared read-only location.

## Implementation

The new flag preserves the existing local content while importing. On every file/directory/symlink the remote tries to write, the importer first checks whether something already exists at that local path. If it does — whether it's a regular file, a symlink (file or directory), or a directory — it's skipped. Non-writable directories under preserved symlinks are also skipped gracefully. All skip decisions are logged to the audit log with a `PRESERVE-LOCAL` prefix.

The setting persists in state, so it survives across resume cycles and delta syncs. During delta sync, previously-skipped files remain protected — the hosting infrastructure symlinks stay safe on every run.

The e2e test builds the realistic hosting structure (symlinked `__wp__` core dir, plugin/theme directory symlinks, drop-in file symlinks, mu-plugin symlinks — all pointing to a read-only shared wordpress directory), runs the import, and verifies every symlink survives initial import, resume, and delta sync.

## Test plan

- [ ] `php -l importer/import.php` passes
- [ ] `cd tests/e2e && npx vitest run tests/import-37-preserve-local.test.js` passes all 6 test groups
- [ ] Existing e2e tests still pass (no regressions)
- [ ] `php import.php files-sync --help` shows the new `--on-docroot-nonempty` option
- [ ] Manual test: run files-sync against a non-empty docroot without the flag → error as before
- [ ] Manual test: run files-sync with `--on-docroot-nonempty=preserve-local` against a directory with hosting symlinks → symlinks preserved, remote files imported around them

### DIFF
--- a/importer/import.php
+++ b/importer/import.php
@@ -1243,12 +1243,14 @@ class ImportClient
     {
         $this->verbose_mode = $options["verbose"] ?? false;
         $this->follow_symlinks = $options["follow_symlinks"] ?? false;
-        $this->docroot_nonempty_behavior = $options["docroot_nonempty_behavior"] ?? 'error';
-        if (!in_array($this->docroot_nonempty_behavior, ['error', 'preserve-local'])) {
-            throw new InvalidArgumentException(
-                "Invalid --on-docroot-nonempty value: {$this->docroot_nonempty_behavior}. " .
-                    "Valid values: error, preserve-local",
-            );
+        if (isset($options["docroot_nonempty_behavior"])) {
+            $this->docroot_nonempty_behavior = $options["docroot_nonempty_behavior"];
+            if (!in_array($this->docroot_nonempty_behavior, ['error', 'preserve-local'])) {
+                throw new InvalidArgumentException(
+                    "Invalid --on-docroot-nonempty value: {$this->docroot_nonempty_behavior}. " .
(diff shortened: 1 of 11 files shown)


importer/import.php
```diff
@@ -4814,6 +4896,22 @@ private function ensure_directory_path(string $dir): void
             $current .= "/" . $part;
 
             if (is_link($current)) {
+                if ($this->docroot_nonempty_behavior === 'preserve-local') {
+                    if (is_dir($current)) {
+                        // Symlink points to a directory — use it as-is, don't remove.
+                        // Check write permissions on the target.
+                        if (!is_writable($current)) {
```
adamziel: Can we trust `is_writable` to cover chmod, user permissions etc?

adamziel: It does seem to cover chmod and uid/gid, but there are still OS layers that may refuse the write (not to mention it can fail). Perhaps this is passable for now, but we may need to revisit at one point.

PR #42, opened 2026-03-11
Add $wpdb-based SQL import for SQLite support

adamziel:
## Summary

The importer can now route SQL through WordPress's `$wpdb` instead of requiring a direct MySQL connection. This enables importing into SQLite databases via the sqlite-database-integration plugin — `$wpdb` handles the MySQL-to-SQLite translation transparently.

Two new integration points, both using the new `--wp-load` flag:

- **`db-apply --wp-load=...`** applies a downloaded `db.sql` through `$wpdb`. The `--target-*` MySQL options become unnecessary.
- **`db-sync --sql-output=wpdb --wp-load=...`** streams SQL directly through `$wpdb` as it downloads, same pattern as `--sql-output=mysql`.

WordPress is loaded with `SHORTINIT` so only the database layer (including the `db.php` drop-in for SQLite) gets bootstrapped — no plugins, themes, or hooks. The `--wp-load` path persists in state for resume support.

Internally, `db-apply` uses an `$exec_query` closure that abstracts the backend, so the streaming loop, resume logic, and URL rewriting work identically for both PDO/MySQL and `$wpdb`. The `db-sync` wpdb mode reuses the same `.sql-buffer` crash-recovery strategy as the mysql mode, splitting the buffer into individual statements via `WP_MySQL_Naive_Query_Stream` since `$wpdb->query()` handles one statement at a time.

## Test plan

- [ ] `db-apply --wp-load=...` with a MySQL-backed WordPress site
- [ ] `db-apply --wp-load=...` with a SQLite-backed WordPress site (sqlite-database-integration)
- [ ] `db-sync --sql-output=wpdb --wp-load=...` streaming mode
- [ ] Resume after interruption preserves `wp_load` state
- [ ] Validation: `--wp-load` with nonexistent path errors clearly
- [ ] Validation: `--sql-output=wpdb` without `--wp-load` errors clearly
- [ ] `db-apply` without `--wp-load` or `--target-*` gives helpful error message
- [ ] Existing `db-apply --target-*` MySQL path still works unchanged

### DIFF
--- a/importer/import.php
+++ b/importer/import.php
@@ -5183,7 +5183,7 @@ class ImportClient
                 }
             }
             // wpdb mode: same buffer cleanup logic as mysql mode.
-            if ($wpdb && !$mysql_conn) {
+            if ($wpdb) {
                 $pending = $sql_buffer;
                 $buffer_file = $this->state_dir . "/.sql-buffer";
                 if ($pending === "" && file_exists($buffer_file)) {
--- a/importer/import.php
+++ b/importer/import.php
@@ -978,6 +978,9 @@ class ImportClient
     /** @var string|null MySQL database for --sql-output=mysql. */
     private $mysql_database;
 
+    /** @var string|null Path to WordPress wp-load.php for $wpdb-based import. */
+    private $wp_load_path;
+
     /** @var resource File descriptor for progress output — STDOUT normally, STDERR in stdout mode. */
     private $progress_fd;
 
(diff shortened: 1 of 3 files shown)


adamziel:
I'll close this. Ideally we wouldn't load WordPress since there may be fatal PHP errors involved in requiring `wp-load.php`. https://github.com/adamziel/streaming-site-migration/pull/43 is a safer, more promising path

PR #43, opened 2026-03-11
Add SQLite support to db-apply

adamziel:
## Summary
- update the sqlite-database-integration submodule to a trunk commit that exposes the PDO-compatible MySQL-on-SQLite driver
- add `db-apply` target-engine support for SQLite via `WP_PDO_MySQL_On_SQLite`, while keeping MySQL as the default path
- document the new CLI options, include the extra driver files in the PHAR build, and add focused SQLite-target E2E coverage

## Testing
- `php -l importer/import.php`
- direct smoke test: apply a handcrafted `db.sql` into SQLite via `php importer/import.php db-apply ... --target-engine=sqlite` and query the resulting database through `WP_PDO_MySQL_On_SQLite`
- `./tests/e2e/node_modules/.bin/vitest run tests/import-38-db-apply-sqlite-target.test.js` *(blocked locally: `ensureSite()` requires passworded `sudo` to manage `/srv/e2e-sites`)*

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,8 @@ On the **migration target** side:
  - ext-json — JSON encoding/decoding
  - ext-hash — hash_hmac, hash_equals
  - ext-zlib — deflate_init/deflate_add for gzip streaming
+ - ext-pdo + ext-pdo_mysql — for MySQL targets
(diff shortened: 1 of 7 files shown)


adamziel:
Flagging this for you @janjakes – I'll try to use the SQLite integration's PDO API to import the data into a SQLite database. Let's see how it works out.

PR #97, opened 2026-04-07
Push model: upload local site to remote WordPress

adamziel:
## Summary

This adds the reverse of the current pull model. Instead of downloading from a remote site, the local machine uploads its database and files to a remote WordPress site with atomic cutover semantics.

The push uses the same multipart protocol, producers, cursors, and HMAC auth as the pull model, just with roles swapped. The receiver is the authority on progress — the client tracks "how much the server consumed," not "how much I sent."

Database tables land with a staging prefix (`_push_wp_*`), then a single `RENAME TABLE` statement atomically swaps all tables. Files land in a staging directory, then swap via symlink or directory rename. A unified commit endpoint orchestrates both swaps — files first (rollbackable), then DB (atomic) — with rollback on partial failure.

Nothing goes live until the explicit commit step. Push-sql and push-files can run in any order, be interrupted, and resume from the receiver's last cursor.

## What's new

**MultipartBodyStream** — streaming multipart/mixed resource backed by `php://temp` for bounded-memory curl uploads with HMAC signing.

**ChunkWriter** — shared filesystem write operations (file chunks, directories, symlinks with security validation) used by the file_receive endpoint.

**Receiver endpoints** — `sql_receive` (rewrites table names to staging prefix, executes SQL), `file_receive` (writes to staging directory), `commit` (validates staging state, swaps files then DB, rolls back on failure).

**CLI commands** — `push-sql`, `push-files`, `push-commit`, all resumable with state management matching the pull model's reentrancy design.

## Test plan

- [ ] PHP syntax checks pass on all 7 files
- [ ] Unit test: MultipartBodyStream round-trip with MultipartStreamParser
- [ ] Unit test: table name rewriting across all SQL statement types
- [ ] Unit test: RENAME TABLE builder handles new, dropped, and existing tables
- [ ] Integration test: push SQL to local receiver, commit, verify tables swapped
- [ ] Integration test: push files to staging dir, commit, verify directory swapped
- [ ] Integration test: commit failure leaves live site untouched
- [ ] E2E: Docker scenario pushing full WordPress site between containers

### DIFF
--- a/packages/streaming-exporter/composer.json
+++ b/packages/streaming-exporter/composer.json
@@ -20,7 +20,9 @@
             "src/class-hmac-client.php",
             "src/class-hmac-server.php",
             "src/class-http-server.php",
-            "src/class-sqlite-driver-pdo.php"
+            "src/class-sqlite-driver-pdo.php",
+            "src/class-multipart-body-stream.php",
+            "src/class-chunk-writer.php"
         ],
         "files": [
             "src/utils.php"
--- a/packages/streaming-exporter/src/class-chunk-writer.php
+++ b/packages/streaming-exporter/src/class-chunk-writer.php
@@ -80,6 +80,18 @@ class ChunkWriter
             $this->file_ctime = $ctime;
         }
 
+        // Resume recovery: if a file was partially written in a previous
+        // request, re-open it in append mode so continuation chunks (where
+        // is_first=false) can still be written.  Without this, the writer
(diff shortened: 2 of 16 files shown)


adamziel:
Superseded by the push design in #315: the database moves as a diff (INSERT/UPDATE/DELETE against baselines) rather than staging-table swaps, and the flow stays local-driven. The file staging + atomic cutover ideas live on in the staged store (#298) and the apply engine (stack step 9).

PR #123, opened 2026-04-13
Fix mkdir failure when SQLite path traverses a VFS symlink

adamziel:
## Summary

Studio's `flat-document-root` creates relative symlinks between WASM VFS mount points — for example, `wp-content → ../docroot/srv/htdocs/wp-content`. The `../docroot` part resolves to the `/docroot` mount inside the VFS, but on the host filesystem it points nowhere. When `db-apply` then tries to `mkdir` the `database/` subdirectory through that symlink, PHP fails with "No such file or directory" even though the recursive flag is set.

The fix walks each path component before `mkdir`, resolving symlinks manually via `is_link()`/`readlink()`. For the example above, `/site/wp-content/database` becomes `/docroot/srv/htdocs/wp-content/database` — a direct path that `mkdir` can create without traversing any cross-mount symlinks.

## Test plan

- [ ] Run `studio site pull-reprint` against a WP.com Atomic site (the reproducer from the issue) and confirm `db-apply` completes without the "Cannot create SQLite directory" error
- [ ] Run `composer test:fast` to verify no regressions in existing unit tests

### DIFF
--- a/packages/reprint-importer/src/import.php
+++ b/packages/reprint-importer/src/import.php
@@ -4567,6 +4567,50 @@ class ImportClient
         return str_replace(';', ';;', $value);
     }
 
+    /**
+     * Walk each component of $path and resolve symlinks encountered
+     * along the way.
+     *
+     * PHP's mkdir() fails when a path component is a relative symlink
+     * whose target only resolves correctly in a VFS namespace — common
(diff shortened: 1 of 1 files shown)


adamziel:
Moving the fix to the Studio app instead — the SQLite path should point directly into the raw docroot, avoiding the VFS symlink entirely.

PR #125, opened 2026-04-13
Fix flat-document-root symlinks broken on host in mounted VFS

adamziel:
## Summary

`flat-document-root` uses `realpath()` to compute relative symlink targets. Inside PHP WASM, `realpath()` returns VFS paths (`/docroot/...`, `/flat/...`), producing symlinks like `../docroot/wp-admin` that resolve inside the VFS but break on the host where the raw directory and site directory live in different trees (e.g. `~/.studio/imports/.../raw` vs `~/Studio/mysite`).

This adds `--host-fs-root` and `--host-flatten-to` options that tell reprint the actual host paths corresponding to the VFS mount points. Symlink targets are computed using these host paths. When not provided, they default to the VFS values (identity mapping, correct for non-WASM usage).

## Test plan

- Added `FlatDocumentRootSymlinkTest` with 4 tests covering identity map, VFS-to-host translation, nested symlinks, and end-to-end host resolution
- All 153 existing import tests pass

### DIFF
--- a/.github/workflows/e2e-playground.yml
+++ b/.github/workflows/e2e-playground.yml
@@ -68,8 +68,13 @@ jobs:
           npx @wp-playground/cli --version || echo "no --version flag"
 
           echo "=== Resolve CLI path ==="
-          CLI_PATH=$(node -e "console.log(require.resolve('@wp-playground/cli/cli.js'))")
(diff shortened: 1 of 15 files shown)


adamziel:
No changes needed in reprint. The fix is in Studio: mount at real host paths instead of synthetic VFS paths, so realpath() and symlinks use real paths from the start.

PR #159, opened 2026-04-24
Batch SQLite inserts in db-apply inside a transaction

adamziel:
## Stacked on #158 (rewriter mapping cache)

After the rewriter stopped re-parsing URL mappings, per-statement PDO exec became the next visible cost in `db-apply`: every INSERT fsyncs the SQLite journal. Under WASM PHP that's ~60 ms of syscall time per statement, and on a realistic wp.com-shaped fixture (57 INSERT batches, 2 rewrite mappings) `exec` was sitting at 3.3 s — roughly a third of db-apply after the rewriter fix.

Wrapping the INSERT loop in one SQLite transaction is the textbook fix. The subtle part is the MySQL dump isn't pure DML.

## The footgun

The dump's header ends with `SET AUTOCOMMIT=0;` and the footer emits a literal `COMMIT;`. When we open a user transaction with `$pdo->beginTransaction()` and the dump's `COMMIT;` statement reaches the sqlite-database-integration driver, the driver translates it faithfully — and closes *our* outer transaction. Our later `$pdo->commit()` then fails with "There is no active transaction".

So we can't wrap the whole dump. We can only batch consecutive DML statements — INSERT / UPDATE / DELETE / REPLACE — and commit the open batch whenever anything else (CREATE TABLE, SET, the trailing COMMIT) comes through.

## Behaviour

* Only the SQLite target uses the batching. Native MySQL autocommit for INSERTs is cheap and the driver quirks around things like `LOCK TABLES` aren't worth the risk.
* The resume cursor is rewound to the last durably-committed statement on shutdown or exception, so we never advertise work as done that was rolled back.
* Every non-DML statement flushes the open batch before it executes, matching MySQL's own "DDL implicitly commits the transaction" semantics and keeping the driver's view of transaction state consistent with ours.

## Tests

All 26 tests across the SQLite target, URL-rewriting, resume-SQL, and SQL-data-fidelity e2e suites pass. The change is transparent on the MySQL target (no code path taken).

## Measured impact (bench numbers pending)

This PR description will be updated with the before/after numbers from the WASM benchmark in `tests/e2e/tests/import-50-db-apply-benchmark.test.js` once the current bench run finishes. Expected: `db-apply` wall drops from ~15 s to ~11 s (~25 %) on the 1000-posts / 3000-postmeta / 2-mapping realistic fixture, extrapolating to ~30 s saved on a real wp.com dump.

### DIFF
--- a/packages/reprint-importer/src/import.php
+++ b/packages/reprint-importer/src/import.php
@@ -5019,6 +5019,25 @@ class ImportClient
             false,
         );
 
+        // Batch statements into SQLite transactions so we fsync once per
+        // cursor-save window (~100 statements) instead of once per INSERT.
+        // The MySQLDumpProducer output never contains BEGIN/COMMIT/LOCK
+        // TABLES, so there is nothing to collide with. We only do this on
+        // the SQLite target; native MySQL autocommit for INSERTs is cheap
+        // compared to SQLite's per-statement journal fsync, and the MySQL
+        // driver is happier without outer transactions around edge cases
+        // like CREATE TABLE statements in the dump.
+        $target_engine = strtolower((string) ($options["target_engine"] ?? "mysql"));
+        $use_transactions = ($target_engine === 'sqlite');
+        $in_transaction = false;
+
+        // Track the last durably-committed checkpoint separately from the
+        // running counters. Rollback on shutdown or failure rewinds the
+        // resume cursor to these values, so we never record statements as
+        // executed when the transaction they lived in was rolled back.
+        $committed_statements_executed = $statements_executed;
+        $committed_bytes_read = $bytes_read;
+
         // Stream db.sql through the query stream and execute
         $query_stream = new \WP_MySQL_Naive_Query_Stream();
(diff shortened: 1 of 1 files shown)


adamziel:
Closing this without merging.

The WASM PHP benchmark (the only runtime that actually matters per the perf brief for this work) does not show a measurable win. Same fixture, back-to-back runs:

- Before this PR (so just #158 landed): 14.82 s db-apply
- After this PR: 15.06 s db-apply

That's inside the ±1 s noise floor we see across WASM runs. The likely reason is that Playground CLI's in-memory emscripten filesystem doesn't pay the per-fsync tax a real SQLite file on disk would, so the transaction batching has nothing to amortise away. Under native PHP on a real disk this fix would almost certainly pay off — the sqlite-database-integration driver does hit the journal on every statement — but we don't have a reliable way to measure that today, so landing it on speculation isn't worth the code it adds.

Leaving the branch around (`adamziel/db-apply-sqlite-tx`) in case someone wants to re-measure under native PHP later.

PR #160, opened 2026-04-24
Skip full MySQL AST build for canonical INSERT statements

adamziel:
## Stacked on #159

Before it touches any URLs, the rewriter parses the enclosing SQL statement so it can hand each value the right content-type hint — `block_markup` for `post_content` / `post_excerpt` / `comment_content`, plain text otherwise. It was doing that by feeding every INSERT through the full `WP_MySQL_Parser`, which inflates a 200 KB grammar into an AST. Under WASM PHP that cost ~125 ms per statement — on a realistic fixture that's ~3 s in `db-apply`, more than the actual URL rewriting it does on each statement.

Dump INSERTs all follow one shape — the exact output `MySQLDumpProducer` emits: `` INSERT INTO `t` (`c1`,`c2`,…) VALUES (…),(…),…; ``. A token walk over the lexer output pulls table, column list, and per-value byte boundaries from that in a fraction of the time, without the grammar build.

## Scope and safety

Anything the fast extractor doesn't recognise falls through to the existing AST-driven parser untouched — UPDATE statements, `INSERT ... SELECT`, `INSERT ... ON DUPLICATE KEY UPDATE`, qualified table names, odd dialects. So for anything other than dump INSERTs, behaviour is byte-identical.

The column map it produces has the same `[start, end, column_name]` shape and the same `[start, end)` byte-offset semantics as the AST path, so callers see an equivalent result regardless of which path ran.

## Tests

All 26 e2e tests across URL-rewriting, SQLite target, SQL-resume, and SQL-data-fidelity pass unchanged. The fast path exercises the common case, the AST path exercises everything else.

## Measured impact

PR description will be updated with benchmark numbers once the current WASM bench run settles — the first run showed roughly flat wall-clock vs #159, well within WASM noise, so the measurable win on this specific fixture is small. The reason to land it anyway is that the AST build is a genuine O(grammar) cost that scales with column count and row count: it matters more on fatter dumps (wpcomsh options blobs, large postmeta values) and under native PHP the saved ~2.5 s becomes more visible.

### DIFF
--- a/packages/reprint-importer/src/lib/url-rewrite/class-sql-statement-rewriter.php
+++ b/packages/reprint-importer/src/lib/url-rewrite/class-sql-statement-rewriter.php
@@ -139,16 +139,223 @@ class SqlStatementRewriter
     }
 
     /**
-     * Parse the SQL with WP_MySQL_Parser to extract the table name and build
-     * an offset→column map. Each map entry is [start, end, column_name] where
-     * start/end are byte offsets of the value expression in the SQL string.
+     * Parse the SQL and build an offset→column map.
      *
-     * Returns null for non-INSERT/UPDATE statements or parse failures — the
-     * caller falls back to auto-detect with no column awareness.
+     * Fast path: a lexer-only walk handles the overwhelming common case
+     * (multi-row `INSERT INTO \`t\` (\`c1\`,\`c2\`,…) VALUES (…),(…),…;` as
+     * emitted by `MySQLDumpProducer`). Every INSERT in a dump matches that
+     * shape, and a full AST build there was costing ~125 ms per statement
+     * under WASM PHP — more than the actual URL rewriting on each statement.
+     *
+     * Slow path: UPDATE and anything the fast path rejects fall through to
+     * the full `WP_MySQL_Parser` AST, which handles complex constructs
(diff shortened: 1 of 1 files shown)


adamziel:
Closing this without merging.

Same story as #159: the WASM PHP benchmark doesn't show a measurable win. On the shared fixture (1000 posts / 3000 postmeta / 60 discovered domains / 2 rewrite mappings / 57 statements), two back-to-back runs after this PR came in at 16.46 s and 15.24 s db-apply wall time, vs 15.06 s for just #159 — i.e. inside noise.

The expectation was that swapping the full `WP_MySQL_Parser` grammar build for a token-only extractor would shave the ~3 s we saw in `stmt_parse_ast` on the realistic profile. That didn't happen under WASM, which suggests the grammar inflation cost either isn't what I thought it was under Playground CLI, or that the lexer walk I substituted has its own per-token overhead that cancels the saved grammar build. Either way I don't have the data to justify the additional ~200 lines of token-walker code.

Under native PHP the grammar is genuinely expensive to inflate (the `.php` file is ~200 KB of compiled tables) and this would probably help, but as with #159 that's speculation until we measure.

Leaving `adamziel/db-apply-light-ast` around for anyone who wants to revisit under native PHP.

PR #163, opened 2026-04-25
Replace the SQL grammar build with a lexer-only column extractor

adamziel:
## What

The URL rewriter parses every INSERT / UPDATE through `WP_MySQL_Parser` just to figure out which column each `FROM_BASE64()` value sits in. The grammar inflates a 200 KB compiled-tables file and runs at ~125 ms per statement under WASM PHP — the largest single chunk of `db-apply` time on a realistic dump.

This PR replaces it with a token-level walker that handles every INSERT and UPDATE shape `MySQLDumpProducer` actually emits, drops the AST path entirely, and removes the grammar dependency.

## How

A single `extract_columns($sql)` walks the lexer output and dispatches on the leading verb:

- **INSERT / REPLACE** — `walk_insert` reads `[modifiers] INTO \`t\` (\`c1\`, \`c2\`, …) [VALUES|VALUE]` then walks each `(…)` or `ROW(…)` tuple, tracking parenthesis depth at the token level. Comma at depth 0 ends a value; closing paren at depth 0 ends the tuple. Optional trailing `;` and `ON DUPLICATE KEY UPDATE …` are recognised.
- **UPDATE** — `walk_update` reads `[modifiers] \`t\` SET \`c1\` = expr [, \`c2\` = expr, …]`. Each expression runs to the next comma at depth 0 or to a clause keyword (`WHERE`, `ORDER`, `LIMIT`, `;`, end of input).
- Any shape neither walker recognises returns null. The rewriter treats null the same as an empty column_map: every `FROM_BASE64()` value falls through to plain-text URL rewriting, which still updates URLs correctly. Only the `block_markup` hint (relevant for ~5 WordPress core columns) is lost on those shapes, and those shapes don't appear in real dumps.

The lexer already handles strings, comments, escaped backticks, hex / binary / null literals, escaped quotes, and so on. The walker only needs to track parenthesis depth at the token level — string-literal tokens that contain `(`, `)`, `,` etc. arrive as a single token and never affect depth. We use `WP_MySQL_Token::get_value()` directly for identifier unquoting (it's SQL-mode-aware where my hand-rolled version wasn't), and `WP_MySQL_Lexer::next_token()` already skips whitespace and comments internally — `significant_tokens` just drops the trailing EOF marker.

`WP_MySQL_Parser`, the static grammar cache, the `parse_insert` / `parse_update` / `extract_identifier` AST helpers, and the `WP_Parser_Node` / `WP_Parser_Grammar` imports all go away.

## Measured impact (WASM PHP, real numbers)

Same-fixture benchmark (1000 posts / 3000 postmeta / 30 options / 60 source domains / 2 rewrite mappings / 57 statements):

| Counter | Baseline (trunk) | This PR | Δ |
|---|---:|---:|---:|
| **db-apply wall** | 15.87 s | **12.10 s** | **−3.77 s, ~24% faster** |
| db-apply PHP-measured total | 10.40 s | 7.22 s | −3.18 s |
| **rewrite phase** | 6.40 s | **3.32 s** | **−3.08 s, ~48% faster** |
| **column extractor (× 24 statements)** | **3.07 s** | **0.220 s** | **~14× faster** |
| exec | 3.47 s | 3.37 s | unchanged |
| parse | 0.52 s | 0.52 s | unchanged |

The walker hits all 24 INSERTs in the bench. Per-statement extraction collapses from ~128 ms to ~9.2 ms.

## Adversarial tests

`tests/UrlRewriting/SqlStatementRewriterLexerWalkerTest.php`, **30 tests**, designed to break a naive paren counter:

- string values containing `(`, `)`, `,`, `;`, the literal text `FROM_BASE64()`, escaped quotes, backticks
- nested `CONVERT(FROM_BASE64('…') USING utf8mb4)` and `IFNULL(NULL, NOW())`
- `REPLACE INTO`, `INSERT LOW_PRIORITY IGNORE INTO`, `VALUES ROW(…)`, singular `VALUE` keyword
- trailing `ON DUPLICATE KEY UPDATE` clause (must stop reading values at `ON`)
- mixed-case keywords, weird whitespace, inline `--` and block `/* */` comments between every token
- `NULL`, `DEFAULT`, `-1`, `0xDEADBEEF` literals
- escaped backticks in identifier names: `` `weird``col` ``
- multibyte UTF-8 in identifiers and values
- SQL-injection-shaped payload as a string value: `'); DROP TABLE wp_users; --` — round-trips byte-identical
- multiple `FROM_BASE64()` values in one row, each mapped to the right column
- direct extractor-engagement tests (via reflection) for canonical INSERT, INSERT without trailing semicolon, single-clause UPDATE, multi-clause UPDATE
- shapes the walker must reject (returns null): INSERT without column list, qualified table name `` `db`.`t` ``, INSERT…SELECT, INSERT…SET — all produce no SQL corruption

All 30 walker tests pass + 170 URL-rewriting unit tests overall + 26 e2e tests across URL rewriting, SQLite target, SQL resume, and SQL data fidelity. **226 tests total, 0 failures.**

### DIFF
--- a/packages/reprint-importer/src/lib/url-rewrite/class-sql-statement-rewriter.php
+++ b/packages/reprint-importer/src/lib/url-rewrite/class-sql-statement-rewriter.php
@@ -97,9 +97,10 @@ class SqlStatementRewriter
             return $sql;
         }
 
-        // Parse the INSERT/UPDATE statement to extract the table name and
-        // build a byte-offset→column map.
-        $parsed_statement = $this->parse_statement($sql);
+        // Recover the table name and a byte-offset→column map from the
+        // INSERT/UPDATE shape so we can hand each FROM_BASE64() value the
+        // right content-type hint downstream.
+        $extracted_columns = $this->extract_columns($sql);
 
         // Iterate over all FROM_BASE64() values using the cursor-based scanner
         $scanner = new Base64ValueScanner($sql);
@@ -118,13 +119,13 @@ class SqlStatementRewriter
 
             // Determine content type hint for this column
             $content_type = null;
-            if ($parsed_statement !== null) {
+            if ($extracted_columns !== null) {
                 $column_name = $this->find_column_at_offset(
-                    $parsed_statement['column_map'],
+                    $extracted_columns['column_map'],
                     $scanner->get_match_offset()
                 );
                 if ($column_name !== null) {
-                    $content_type = $this->get_content_type($parsed_statement['table'], $column_name);
+                    $content_type = $this->get_content_type($extracted_columns['table'], $column_name);
                 }
             }
 
@@ -176,7 +177,7 @@ class SqlStatementRewriter
      *
      * @return array{table: string, column_map: list<array{int, int, string}>}|null
      */
-    private function parse_statement(string $sql): ?array
+    private function extract_columns(string $sql): ?array
     {
         $tokens = self::significant_tokens($sql);
         $token_count = count($tokens);
--- a/packages/reprint-importer/src/lib/url-rewrite/class-sql-statement-rewriter.php
+++ b/packages/reprint-importer/src/lib/url-rewrite/class-sql-statement-rewriter.php
@@ -241,7 +241,10 @@ class SqlStatementRewriter
         if ($cursor >= $token_count) {
             return null;
         }
-        $table_name = self::unquote_identifier_token($tokens[$cursor]);
+        $table_token_id = $tokens[$cursor]->id;
+        $table_name = ($table_token_id === WP_MySQL_Lexer::BACK_TICK_QUOTED_ID || $table_token_id === WP_MySQL_Lexer::IDENTIFIER)
+            ? $tokens[$cursor]->get_value()
+            : null;
         if ($table_name === null) {
             return null;
         }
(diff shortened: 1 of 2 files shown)


adamziel:
## Real WASM bench numbers (replacing the projection in the PR description)

Caught a bug while measuring: the lexer's final EOF token (id -1) was tripping the trailing-tokens guard, so the walker was returning null on every real-dump INSERT and silently falling through to the AST path. None of the 26 indirect tests caught it — they only checked end-to-end output, which the AST produced equally well. Fixed by filtering EOF in `significant_tokens` and adding two direct fast-path engagement tests via reflection that would have flagged the regression loudly.

### Measured impact, same fixture, same WASM PHP runtime

| Counter | Baseline (trunk + profile only) | + this PR | Δ |
|---|---:|---:|---:|
| **db-apply wall time** | **15.87 s** | **12.93 s** | **−2.94 s, ~19% faster** |
| db-apply PHP-measured total | 10.40 s | 7.62 s | −2.78 s |
| **rewrite phase** | **6.40 s** | **3.56 s** | **−2.84 s, ~44% faster** |
| **stmt_parse_ast (× 24 statements)** | **3.07 s** | **0.232 s** | **~13× faster** |
| exec | 3.47 s | 3.53 s | unchanged |
| parse | 0.52 s | 0.53 s | unchanged |

The walker hits all 24 canonical INSERTs on the bench fixture; AST cost per statement collapses from 128 ms to ~9.7 ms. Real numbers vindicate the projection.

Test count: 28 walker unit + 165 existing unit + 26 e2e = **219 tests pass**.

PR #185, opened 2026-04-29
Reprint wizard: hand off to Playground via Blueprint

adamziel:
## What it does

Ships a deployable WordPress.com → Playground import wizard. The user signs in with WordPress.com, picks a site, and clicks **Start import**; the wizard rotates a fresh export secret, hands the import off to a new WordPress Playground tab via Blueprint URL, and the cloned site renders entirely in the browser without doubled scope prefixes, missing-asset 404s, or redirect loops on uploads.

## Rationale

The wizard exists to bridge two things that don't fit together cleanly:

- **WordPress.com OAuth.** Required to authorise `reprint-exporter` on the source site. Has to live on a domain WordPress.com will redirect back to.
- **`reprint.phar pull`.** Wants megabytes of memory, long-running connections, and direct SQLite access. None of which fit comfortably in the same browser tab as an OAuth UI.

Splitting them lets each side keep what it needs. The wizard owns identity; Playground owns the heavy work; nothing is shared between them except the per-import secret and the export endpoint URL, both passed in the URL fragment so they never reach a server log.

## Implementation

**`reprint-ui/index.php`** — the deployed wizard. Three steps:

1. WordPress.com OAuth (`oauth-redirect.php` is the callback).
2. Site picker, then **Start import**: the wizard provisions the exporter on the chosen site and rotates a fresh secret.
3. Hands off to `https://playground.wordpress.net/?blueprint-url=…#api=…&secret=…`.

**`reprint-ui/blueprint.json` (and `?action=blueprint`)** — Playground reads this and:

- Drops `reprint.phar` and `reprint-import.php` into the WASM filesystem.
- Defines `DB_NAME`, `ABSPATH`, `WP_CONTENT_DIR` in a preload step so Atomic `wp-config.php`s don't blow up on missing host-injected DB credentials.
- Lands on `/reprint-import.php`.

**`reprint-ui/reprint-import.php`** — the page that runs inside Playground:

- Streams `reprint.phar pull` as NDJSON; activation runs in the same request (a separate POST gets 302'd by Playground's request handler before any of our code runs).
- Hard-fails if the imported SQLite is empty (used to silently land on the WP install wizard).
- Drops a mu-plugin (`0-reprint-playground-glue.php`) that:
  - Streams missing `/wp-content/uploads/*` from the source via `CURLOPT_WRITEFUNCTION` (a 302 to the source would loop through Playground's service worker until `ERR_TOO_MANY_REDIRECTS`).
  - Disables Jetpack's frontend CSS imploder (same family of asset bundling that needs Atomic-side route handlers).

**Importer changes shipped alongside:**

- `EXIT_AFTER_IMPORT` define lets web-SAPI embedders run cleanup after `pull` returns instead of `exit()`-ing the process. Activation now lives in the same `try` scope as the include, so exceptions surface as `{type:'error'}` events.
- `deactivate_path_incompatible_plugins()` runs at the end of `db-apply` and removes `page-optimize` from `active_plugins` when `--new-site-url` has a non-`/` path. Skipped when the path is empty or `/`.
- `REPRINT_INSECURE_TLS=1` escape hatch — Playground's `php-wasm` fetch shim doesn't verify TLS the same way curl does, and the user's browser is the actual transport. Threat-modelled in a comment block at the call site.
- `reprint_apply_curl_ca_bundle()` mirrors PHP's `openssl.cafile` ini value onto `CURLOPT_CAINFO` per handle. No-ops when `curl.cainfo` is set in `php.ini` (host already configured curl) or `openssl.cafile` is empty (no information to pass through). Tagged `TODO` pointing at `WordPress/wordpress-playground` for the upstream fix.

**CI:** every workflow that runs `tests/e2e/ci/setup-infrastructure.sh` now installs PHP via `shivammathur/setup-php@v2`. The script hard-errors if PHP isn't on PATH after the action runs — no fallback to `add-apt-repository ppa:ondrej/php`, which has had multiple Launchpad-side outages this week.

**Known follow-ups:**

- #197 — move the wizard's Atomic versioned-plugin/theme layout collapse out of `run_local_activation()` and into the importer via a `RuntimeManifest`-declared `auto_prepend_file`.

## Testing instructions

- [ ] Sign in to wp.com on the deployed wizard at `https://adorable-encouraging-triton.wpcloudstation.dev/`, pick an Atomic site, click **Start import**.
- [ ] The tab should navigate to `https://playground.wordpress.net/?blueprint-url=…#api=…&secret=…`.
- [ ] The importer page loads, shows the **Don't navigate away** warning, and progress phases stream live.
- [ ] When the import finishes, click **Open the imported site →** — the home page should render the cloned site without doubled `/scope:<slug>/` paths and without 404s on stylesheets.
- [ ] Open a post that references a media item that wasn't downloaded yet — the image should render (proxied from the source) without the iframe redirect-looping.
- [ ] Locally: `cd tests/e2e && ./run-all-tests.sh` — PHP 7.4 through 8.5 should all set up cleanly via `shivammathur/setup-php@v2`.


adamziel:
## Code review (self) — `/code-review-adam` pass

Run against the diff at HEAD (`7b2bc0f9`).

**Streaming broken in disguise** — the upload proxy in `0-reprint-playground-glue.php` uses `CURLOPT_RETURNTRANSFER => true` then `echo $body`, which buffers the whole upload body into PHP memory before a byte hits the browser. For a 100MB video that's 100MB of WASM PHP memory. The ask: stream via `CURLOPT_WRITEFUNCTION` (or `CURLOPT_FILE => fopen('php://output','w')`) with `CURLOPT_RETURNTRANSFER => false`, emit headers from the writefunction's first invocation.

**Silent failure with misleading reassurance** — `run_local_activation()` line 542 catches `Throwable` from the page-optimize deactivation block with `/* non-fatal — surface via row_counts */`, but `row_counts` is the wp_options/wp_users probe, completely unrelated. If the UPDATE throws, we report `ok=true` and the user lands on a site whose page-optimize is still active. The ask: append the exception to the activate event payload, or hard-fail.

**Leaked platform-specific knowledge** — the Atomic versioned-plugin collapse (reprint-import.php:570-630) lives in the wizard's activation step, but it's a transformation on the layout `flat-docroot` produced. flat-docroot is the layer that owns "make this WP-bootable on the target host"; the wizard now silently knows about Atomic's versioned plugin/theme tree. The ask: move into `flat-docroot` itself, or into a reusable helper in the importer package.

**try/finally for resource teardown missing** — the upload proxy's `curl_init` / `curl_close` pair has no `finally`. If the HEADERFUNCTION callback throws, the handle leaks. The ask: `try { … } finally { curl_close($ch); }`.

**Imprecise name** — `_reprint_self_origin()` returns the full URL with path (`https://playground.wordpress.net/scope:foo`), but "origin" in URL spec means scheme + host + port. The ask: rename to `_reprint_canonical_site_url()` so the call site `--new-site-url=` reads correctly.

**Duplicate PDO connections** — lines 525 and 554 each open a fresh `new PDO('sqlite:' . $dst)` against the same file within the same function. Single source of truth — open once, share.

**Hardcoded constants without a named source** — preload paths (`/wordpress/wp-content/plugins` etc.) are correct because Playground's documentRoot is `/wordpress`, but that's a Playground convention. The ask: at least name the assumption in a comment, ideally lift to a single constant set.

### Trade-offs articulated well (no findings)

- `REPRINT_INSECURE_TLS=1` — threat model named precisely; user's browser is the actual transport; rotating 60-min HMAC; comment block in the helper makes it impossible to silently spread.
- `unset($wpdb) + define('WP_CLI', true)` — uses the SQLite driver's documented WP-CLI escape valve; comment explains both the bypass and the inverse stdClass-stub trap that crashes inside wp-admin/schema.php.

### Adversarial pass

The phar's CLI entry point `exit(0)`s after `pull` succeeds, and `stream_pull_and_activate` works around that by registering activation as a `register_shutdown_function`. Structurally fragile: any future phar change that throws or moves cleanup will silently re-shape what the activation event looks like, and the JS's "if !activateResult" branch will lie when the real failure is upstream. The phar already exposes an `IMPORTER_WEB_ENTRY` define that neuters `start_server` — the right fix is a peer (`IMPORTER_NO_EXIT`, or have `IMPORTER_WEB_ENTRY` imply no-exit) that returns control to the caller after `pull`. Activation then runs in the `try` block where exceptions surface as `{type:'error'}` events instead of in a shutdown handler with no exception channel.

---

Stacked PR with the fixes incoming.

PR #186, opened 2026-04-30
Reprint wizard: code-review fixes (stacked on #185)

adamziel:
Stacked on #185. Applies the findings from the `/code-review-adam` self-review posted on that PR.

## What changes

1. **importer: `IMPORTER_NO_EXIT`** — `import.php`'s CLI dispatcher used to `exit($exit_code)` at the end of `pull`. The wizard worked around that with `register_shutdown_function`, but a shutdown handler has no exception channel: anything that throws in activation just disappeared instead of becoming a `{type:'error'}` ndjson event. The dispatcher now stashes the exit code on `$GLOBALS['REPRINT_IMPORTER_EXIT_CODE']` and only `exit()`s when `IMPORTER_NO_EXIT` is undefined / falsy. The wizard defines it before `include $phar` and runs activation in the same `try/catch`.

2. **importer: flat-docroot collapses Atomic versioned plugin/theme dirs** — Atomic ships shared plugins as `/wordpress/plugins/<plugin>/<version>/`, but WordPress looks at `/wp-content/plugins/<plugin>/<plugin>.php` without the version segment. The collapse used to live in the wizard; moved into `ImportClient::flatten_collapse_versioned_dirs()` so any importer (Studio's pull-reprint, e2e fixtures) gets a WP-bootable layout for free.

3. **wizard: streaming upload proxy** — the old proxy did `CURLOPT_RETURNTRANSFER => true; echo $body`, buffering whole uploads into WASM PHP memory before a byte hit the wire. Now uses `CURLOPT_WRITEFUNCTION` to flush each libcurl chunk straight to `php://output`, with `try/finally` around the handle.

4. **wizard: surface page-optimize deactivation failures** — the previous catch swallowed `Throwable` with a comment claiming "non-fatal — surface via row_counts", but `row_counts` is unrelated. Activation now appends to a `warnings` array; the JS logs each warning so the user sees what we recovered from.

5. **wizard: rename `_reprint_self_origin` → `_reprint_canonical_site_url`** — "origin" means scheme+host+port; the function returns the full URL including the `/scope:<slug>` path. New name reads correctly at the call site.

6. **wizard: share one PDO handle in `run_local_activation`** — the deactivation UPDATE and the row-count probe each opened their own connection. Consolidated to a single `$pdo` so future PRAGMA / function-registration setup lives in one place.

## Test plan

- [ ] CI green: PHPUnit (PHP 8.1-8.4), E2E PHP matrix, Playground CLI
- [ ] Re-run the wizard against an Atomic source — verify a) flat-docroot's audit log shows `Collapsed versioned dir` entries; b) the importer's pull command exits cleanly without slamming the wizard's stack; c) activation warnings (if any) appear in the wizard's log box; d) a 100MB+ upload streams through the proxy without OOMing.

### DIFF
--- a/packages/reprint-importer/src/import.php
+++ b/packages/reprint-importer/src/import.php
@@ -4529,12 +4529,30 @@ class ImportClient
             }
         }
 
+        // Phase 3: Collapse Atomic-style versioned plugin/theme dirs.
+        //
+        // Atomic ships shared plugins/themes under a versioned subdir:
+        //   /wordpress/plugins/jetpack/15.8-a.7/jetpack.php
+        //   /wordpress/themes/twentytwentyfour/2.4/style.css
+        // WordPress, however, looks up entry points without the version
+        // segment — /wp-content/plugins/jetpack/jetpack.php — so the
+        // imported tree on its own gives WP nothing to load and every
+        // .../static/* asset 404s. After Phase 2 has placed the
+        // plugins/themes/mu-plugins symlinks into the flattened
+        // wp-content, scan the top-level entries: when one resolves to
+        // a directory containing exactly one child that looks like a
+        // version (digits and dots, optionally with a hyphenated
+        // suffix), retarget the symlink one level deeper so WP sees
+        // the entry-point file at the expected path.
+        $collapsed = $this->flatten_collapse_versioned_dirs($flatten_to);
+
         $this->audit_log(
             sprintf(
-                "FLAT-DOCUMENT-ROOT | Complete: %d created, %d refreshed, %d force-replaced",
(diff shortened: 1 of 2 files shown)


adamziel:
Superseded by stacked PRs #188 → #189 → #190 → #191. Atomic versioned-plugin collapse fix dropped pending a proper repro fixture — the existing e2e Atomic coverage doesn't exercise the versioned-subdir layout, and Studio / earlier Playground imports succeeded going through apply-runtime, so the symptom I observed against improdotblog/adamadam.blog deserves a tighter diagnosis before landing.

PR #196, opened 2026-05-01
Download public uploads outside WordPress when possible

adamziel:
## What it does

Downloads mapped public `wp-content/uploads/` files directly from the source site's public upload URL when the path can be mapped from `paths_urls.uploads.basedir` to `paths_urls.uploads.baseurl`.

Interrupted public-upload downloads are resumable in both cases:

- If the public server supports HTTP Range, the importer resumes the `.reprint-public-download` temp file with `Range: bytes=<offset>-` and requires HTTP `206`.
- If the public server does not support HTTP Range, the importer validates the existing temp prefix against PHP with a per-file `md5` prefix hash, moves that prefix into the final local path, and resumes `file_fetch` from the same byte offset.

Falls back to normal `file_fetch` whenever the public path is unavailable, unmapped, returns the wrong status/size, cannot be validated, or cannot be resumed safely.

Stack: based on #195. The perf job compares this PR against #195's branch, so the reported delta isolates the direct-upload change.

## Rationale

Large media files spend too much time moving through WordPress/PHP when they are already publicly reachable through the source site's upload URLs. Pulling those files over plain HTTP avoids PHP reading, chunking, and re-streaming every large JPEG/PDF/ZIP.

The public URL is still only the happy path. `file_fetch` remains the correctness path for private uploads, unmapped paths, unsupported public responses, validation failures, and resumed transfers that cannot continue through HTTP Range.

## Implementation

Adds a public-upload fast path while building `file_fetch` batches:

- Maps upload paths to public URLs from preflight `paths_urls.uploads`.
- Downloads mapped files directly to a deterministic `.reprint-public-download` temp file.
- Persists `state["public_upload"]` so an interrupted process can resume the public transfer.
- Uses cURL resume support for HTTP Range and requires `206` on resumed requests.
- Adds a lightweight exporter `file_prefix_hash` endpoint for one-file `md5` prefix validation.
- Synthesizes a `file_fetch` cursor for the validated interrupted file so PHP streaming starts at the same byte offset.
- Keeps failed or unvalidated public transfers on the existing `file_fetch` fallback path.

Adds the `files-pull-public-uploads` benchmark with a dedicated public-upload benchmark site. The timed large transfer is the upload file itself. CI requires the PR run to prove `direct_upload=yes`; the trunk/#195 baseline is allowed to report `direct_upload=no`.

Latest completed Performance Report on commit `b523ff96`, using the isolated 24 MiB public-upload benchmark:

- PR: `2.92 s`, `direct_upload=yes`, `public_fallback=no`
- trunk/#195 baseline: `5.27 s`, `direct_upload=no`, `public_fallback=no`
- delta: `-2.35 s` / `-44.6%`

## Testing instructions

```bash
php -l packages/reprint-exporter/src/export.php
php -l packages/reprint-exporter/src/class-http-server.php
php -l tests/Import/PublicUploadDownloadTest.php
php -l tests/Import/FileBodyStreamingTest.php
vendor/bin/phpunit --bootstrap tests/bootstrap.php tests/Import/PublicUploadDownloadTest.php --testdox
vendor/bin/phpunit --bootstrap tests/bootstrap.php tests/Import/FileBodyStreamingTest.php --testdox
composer analyze
git diff --check
```

### DIFF
--- a/.github/workflows/perf-report.yml
+++ b/.github/workflows/perf-report.yml
@@ -138,13 +138,14 @@ jobs:
           BENCH_JSON_OUT: bench-pr.json
           BENCH_MD_OUT: bench-pr.md
           BENCH_STAGES: ${{ steps.stages.outputs.stages }}
+          BENCH_REQUIRE_DIRECT_PUBLIC_UPLOAD: '1'
           E2E_EXPORTER_PROJECT_ROOT: ${{ github.workspace }}
         run: node benchmark/bench-pull.mjs
 
       - name: Reset benchmark sites before trunk
         run: |
-          sudo rm -rf /srv/e2e-sites/large-directory /srv/e2e-sites/large-single-file
-          sudo rm -f /tmp/e2e-site-large-directory.lock /tmp/e2e-site-large-single-file.lock
+          sudo rm -rf /srv/e2e-sites/large-directory /srv/e2e-sites/large-single-file /srv/e2e-sites/public-upload-bench
+          sudo rm -f /tmp/e2e-site-large-directory.lock /tmp/e2e-site-large-single-file.lock /tmp/e2e-site-public-upload-bench.lock
 
       - name: Bench trunk
         working-directory: tests/e2e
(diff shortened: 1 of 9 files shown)


adamziel:
this doesn't seem to be speeding things up in a meaningful way

PR #207, opened 2026-05-09
ci(perf): add file-fetch-mixed-batch scenario + restore trunk-exporter baseline

adamziel:
**Stacked on #205.** Reviewers: this PR's diff vs base only shows the bench/workflow changes; the heuristic flip itself is in #205.

## Why the perf comment can't see #205 today

Both the PR and trunk bench steps in `perf-report.yml` deploy the **PR's** exporter to the test site (because `E2E_EXPORTER_PROJECT_ROOT` defaults to `${{ github.workspace }}` on both sides). So whatever server-side perf change a PR makes, the trunk bench is hitting the same code as the PR bench, and the diff only captures importer-side changes. #205 is a server-side change → invisible.

The bench is also limited to two SQLite stages by `focused-stages.txt`, neither of which exercises file_fetch.

## What this PR does

1. **Restore the trunk-exporter side-by-side checkout** that commit e9d740d simplified away.
   - New `Checkout trunk exporter baseline` step → `trunk-exporter/`
   - New `Install trunk exporter dependencies` step (composer install on the side checkout)
   - Trunk bench step's `E2E_EXPORTER_PROJECT_ROOT` now points at `${{ github.workspace }}/trunk-exporter`

   Why e9d740d's intent isn't violated: that commit was concerned with the Playground SQLite stages (`db-pull` / `db-apply`), which exercise the SQL stream — no file_fetch. The dual-checkout matters for file-fetch perf only.

2. **New `file-fetch-mixed-batch` bench scenario.** Provisions a fixture inside the existing `large-single-file` site at `test-data/bench-mixed-batch/`:
   - 100 × 1 KB CSS files (highly repetitive, gzips to ~80–90× ratio)
   - 5 × 20 KB PNG-shaped files (random bytes, incompressible)

   Issues a single `file_fetch` POST and reports `encoding`, `response` (total wire bytes), and multipart part counts in `details`. The renderer prints those side-by-side, so the trunk-vs-PR delta is legible at a glance.

3. **Add the new stage to `focused-stages.txt`** so it actually runs on PR perf checks. Also cleaned up the stale comment in that file (it was written from a previous PR's narrative about loading the Rust parser only on the PR side; that's no longer the situation post-#206).

## Expected perf comment shape after #205 lands

| Stage | PR | trunk | Δ | Details (PR / trunk) |
| --- | ---: | ---: | --- | --- |
| `file-fetch-mixed-batch` | small | larger | wire bytes shrink ~50 % | encoding=gzip ; encoding=identity |

In-process measurement on a synthetic mixed fixture (200 CSS + 5 PNG): trunk 406 KB → PR 207 KB on the wire (-49 %). The CI fixture is calibrated to the same shape so the same magnitude should show up.

## Test plan

- [ ] CI: PR-side bench should show `encoding=gzip`, smaller `response` bytes
- [ ] CI: trunk-side bench should show `encoding=identity`, larger `response` bytes
- [ ] CI: SQLite stages still complete in budget (the manifest fix from #206 is preserved by the rebase)
- [ ] Sticky perf comment refreshes with the file-fetch-mixed-batch row

### DIFF
--- a/.github/workflows/perf-report.yml
+++ b/.github/workflows/perf-report.yml
@@ -64,6 +64,19 @@ jobs:
         with:
           submodules: true
 
+      # Side-by-side checkout of the trunk exporter so the trunk bench step
+      # serves test-site requests against trunk's exporter code, not the
+      # PR's. Without this, both bench steps would deploy the PR's exporter
+      # (E2E_EXPORTER_PROJECT_ROOT defaults to the workspace), and the
+      # trunk-vs-PR diff would only ever capture importer-side changes —
+      # any exporter-side perf change in the PR would be invisible.
+      - name: Checkout trunk exporter baseline
+        uses: actions/checkout@v5
+        with:
+          ref: trunk
+          path: trunk-exporter
+          submodules: true
+
       - name: Resolve focused bench stages
         id: stages
         run: |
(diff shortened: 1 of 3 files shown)


adamziel:
Closing — #205 has landed on trunk. The dual-checkout + mixed-batch scenario in this PR validated #205's effect (wire 269.1 KiB → 102.8 KiB, −62 %), but the bench scenario itself is held for now per the discussion thread; it can be revived as a fresh PR if/when we want it permanently in the perf gate.

PR #209, opened 2026-05-11
perf(exporter): ship a must-use plugin loader to skip WP boot tail for API requests

adamziel:
## Why

The regular plugin at `wp-content/plugins/reprint-exporter-wp/index.php` intercepts during the same plugin-load step that fires for every other active plugin. By the time it runs, `wp-settings.php` has already:

- parsed `wp-config.php` and connected `\$wpdb` to MySQL
- enumerated and required every file in `wp-content/mu-plugins/`
- enumerated and required every active regular plugin that sorts alphabetically before `reprint-exporter-wp`

On wp.com Atomic / large managed-host sites those add up to **50–200 ms per request** — paid on every `file_fetch` / `file_index` / SQL batch in a sync, which adds up over a multi-thousand-batch migration.

## What

`reprint-exporter-wp/mu-plugin/0-reprint-exporter.php` — a small must-use plugin loader (84 lines, mostly comments) that users copy or symlink into `wp-content/mu-plugins/`.

The leading `0-` makes it sort first in `wp_get_mu_plugins()`'s alphabetical scan. On API requests it requires `lib.php` from the regular plugin's directory, dispatches, and `exit`s before:

- any other mu-plugin loads
- any regular plugin loads
- everything `wp-settings.php` does after `wp_load_mu_plugins()` (themes, locale, query, rewrite)

On non-API requests it returns silently and normal WP boot proceeds untouched.

What's still paid for an API request: `wp-config.php` + `wp-load.php`, the early section of `wp-settings.php`, `\$wpdb` instantiation (which the export endpoints need anyway), and any mu-plugin that sorts before `0-reprint-exporter` (none, by construction). What's bypassed: the rest of mu-plugin loading + every regular plugin + the rest of `wp-settings.php`.

## Failure modes (all silent, never fatal)

A fatal mu-plugin would take the whole site offline for every visitor. This loader **never** fatal-errors:

- Regular plugin directory missing or unreadable → no-op return, WP serves a 404 for the unhandled API URL.
- `WP_CONTENT_DIR` undefined (e.g. someone wires us via `auto_prepend_file` before `wp-config.php` runs) → no-op return.
- Empty-string value for the GET param (some HTTP clients drop the `=value`) → `isset()` still triggers, intercept fires.

Safe to install alongside the regular plugin. On API requests the mu-plugin exits before the regular plugin would have loaded; on non-API requests the mu-plugin no-ops and the regular plugin continues to serve its admin-UI responsibilities.

## Tests

`tests/MuPluginLoaderTest.php` — 9 tests, 38 assertions, all green. Each runs the loader inside a subprocess so its `exit()` doesn't bring PHPUnit down.

Positive (intercept + exit):
- `?reprint-api` triggers handler
- legacy `?site-export-api` still recognised
- empty-string param value (`?reprint-api=`) still triggers (`isset` semantics)
- both params present + arbitrary other params present still triggers

Negative (silent no-op return):
- no API param present → no intercept, no spurious lib require
- empty `\$_GET` → no intercept
- non-API page view does **not** require `lib.php` (verified by dropping a `lib.php` that throws on load — if the loader ever requires it for a non-API request, the test would see the throw)

Failure-tolerance:
- missing `lib.php` (plugin dir gone) → no-op return, no PHP errors
- `WP_CONTENT_DIR` undefined → no-op return, no PHP errors

## Install (from the README update)

```bash
# Symlink (recommended — picks up plugin updates automatically):
ln -s ../plugins/reprint-exporter-wp/mu-plugin/0-reprint-exporter.php \
      wp-content/mu-plugins/0-reprint-exporter.php

# Or copy (simpler, won't track upgrades):
cp wp-content/plugins/reprint-exporter-wp/mu-plugin/0-reprint-exporter.php \
   wp-content/mu-plugins/0-reprint-exporter.php
```

## Test plan

- [x] `phpunit MuPluginLoaderTest.php` — 9 / 9 green
- [ ] CI green

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -16,6 +16,13 @@ parameters:
         analyseAndScan:
             - reprint-exporter-wp/wordpress/
             - reprint-exporter-wp/index.php
+            # mu-plugin/install.php is included only from activation /
+            # deactivation hooks, where WP is fully booted. It calls
+            # update_option / get_option / delete_option / wp_mkdir_p
+            # unconditionally because that context guarantees them —
+            # but PHPStan can't see that. Same exclusion strategy as
+            # index.php above.
+            - reprint-exporter-wp/mu-plugin/install.php
             - reprint-exporter-wp/vendor/
             - lib/sqlite-database-integration/packages/mysql-on-sqlite/src/mysql/native/
         analyse:
--- a/reprint-exporter-wp/README.md
+++ b/reprint-exporter-wp/README.md
@@ -22,6 +22,45 @@ When a request arrives at `https://example.com/?reprint-api` (or the legacy `?si
 
 This gives us a clean execution environment while using WordPress's front controller as the entry point.
 
+### Optional fast path: install as a must-use plugin
+
+The regular plugin still pays for everything wp-settings.php does
+before it fires plugin-load: `$wpdb` instantiation, mu-plugin loading,
+and any regular plugins that sort alphabetically before
+`reprint-exporter-wp`. On a typical wp.com Atomic / large managed-host
+site that's 50–200 ms per request — paid on every file_fetch,
+file_index, and sql batch in a sync, which adds up over a
+multi-thousand-batch migration.
+
+To skip most of that, install the bundled mu-plugin loader. Either copy
+or symlink it into `wp-content/mu-plugins/`:
+
+```bash
+# Copy (simpler, but won't track plugin upgrades):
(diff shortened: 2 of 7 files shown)


adamziel:
Closing — the measurement shows the win is real but smaller than originally framed (~3-10 ms per cold-worker request on plugin-heavy sites, undercounted vs real wp.com Atomic boot cost). Holding off pending a decision on whether the install/uninstall + admin-notice complexity is worth shipping for that magnitude, or whether to invest the effort in a different optimization.

PR #210, opened 2026-05-11
[codex] Add aggressive transfer profile

adamziel:
## Summary

Adds an opt-in `--transfer-profile=aggressive` mode for sustained sequential transfers and fixes tuner state overrides so explicit CLI/profile start values replace persisted preflight state.

The aggressive profile:

- removes intentional duty-cycle sleep (`duty=1`)
- raises the per-request execution budget to 30s
- starts file chunks at 16 MiB
- starts file-index batches at 50k entries

Default behavior remains the current balanced profile.

## Benchmarks

Local e2e source, PHP 8.2.29, file transfer only. Each row runs default preflight, then times `files-pull`.

| Dataset | Default | Aggressive profile | Delta | Notes |
|---|---:|---:|---:|---|
| Favorable: `large-directory` expanded to 220 MiB so default hits a partial response | 11.741s, 2 attempts, 5.000s tuner sleep | 9.252s, 1 attempt, 0s sleep | 1.27x faster | profile avoids the forced second request and duty sleep |
| Unfavorable: existing 92 MiB `large-directory`, completes in one default request | 5.673s, 1 attempt, 0s sleep | 6.385s, 1 attempt, 0s sleep | 12.6% slower | no sleep to remove; larger chunk framing is not beneficial here |

## Validation

- `php -l packages/reprint-importer/src/import.php`
- `php -l tests/Import/AdaptiveTunerTest.php`
- `vendor/bin/phpunit tests/Import/AdaptiveTunerTest.php --colors=never`

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,6 +203,17 @@ The command returns one of three exit codes:
 
 Which is to say, you'll need to wrap it in a loop that runs until failure or full completion.
 
+**Transfer profile**
+
+By default, RePrint uses the `balanced` transfer profile: short requests with an intentional duty-cycle pause between partial requests. On trusted sources that can tolerate sustained sequential transfer load, use `--transfer-profile=aggressive` to remove the duty-cycle pause and use larger request budgets:
+
+```bash
+php reprint.phar files-pull "$URL" --state-dir="$STATE_DIR" --fs-root="$FS_ROOT" --secret="$SECRET" \
+    --transfer-profile=aggressive
+```
+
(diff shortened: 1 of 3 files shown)


adamziel:
Closing for this db-apply speedup pass. This transfer-profile experiment targets file transfer and even records an unfavorable slower case; it is not a surviving database apply phase speedup. The surviving db-apply speedups are #235 and #239.

PR #211, opened 2026-05-11
[codex] Batch file fetch state saves

adamziel:
## Summary

Reduces hot-path state persistence during `files-pull`:

- completed file parts are checkpointed in batches instead of forcing one state write per file
- unfinished file parts still checkpoint immediately so mid-file resume keeps its byte counter
- the index count used only for `SAVE CURSOR` audit logging is cached per index version, avoiding repeated full reads of `.import-index.jsonl`

This keeps the transfer sequential and preserves the existing periodic checkpoint behavior.

## Benchmarks

Local e2e source, PHP 8.2.29, `files-pull` only. Both trunk and this branch used the same high-throughput tuning flags to isolate state-save overhead:

`--duty=1 --max-exec=60 --file-chunk-start=16777216 --file-chunk-max=16777216 --index-batch-start=50000 --index-batch-max=50000`

| Dataset | Trunk | This branch | Delta | State writes |
|---|---:|---:|---:|---:|
| Less favorable mixed fixture: 5,638 files, 220 MiB | 5.740s | 3.262s | 1.76x faster | 5,687 -> 173 |
| Favorable many-small-file fixture: 25,638 files, 299 MiB on disk | 26.344s | 11.013s | 2.39x faster | 25,787 -> 673 |

## Validation

- `php -l packages/reprint-importer/src/import.php`
- `php -l tests/Import/FileBodyStreamingTest.php`
- `vendor/bin/phpunit tests/Import/FileBodyStreamingTest.php --colors=never`

I also tried the targeted e2e mid-file-resume test locally, but that run was not usable as validation because the test hook crash left the local source server unreachable before the resume assertions.

### DIFF
--- a/packages/reprint-importer/src/import.php
+++ b/packages/reprint-importer/src/import.php
@@ -1052,6 +1052,9 @@ class ImportClient
     /** @var int Running count of files imported in the current invocation. */
     private $files_imported = 0;
 
+    /** @var int|null Cached index count for repeated save-state audit logs. */
+    private $save_state_index_count = null;
+
     /** @var int|null Total entries in the current download list.  Set once
      *  at the start of download_files_from_list() by counting newlines. */
     private $download_list_total = null;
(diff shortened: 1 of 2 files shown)


adamziel:
Closing for this db-apply speedup pass. This targets files-pull state saves, not the database apply phase required by the current evaluation. The surviving db-apply speedups are #235 and #239, both verified with full playground-sqlite-db-apply timings.

PR #212, opened 2026-05-11
[codex] Use identity for mixed file fetch compression

adamziel:
## Summary

Makes `file_fetch` use identity encoding for mixed text/binary path lists. Text-only responses still gzip, and binary-only responses still stay identity.

This prevents one compressible file from forcing response-level gzip over large already-compressed media payloads. The importer remains sequential and unchanged in this PR; a naive importer-side path-list splitter was tested locally but rejected because it created too many sequential fetch requests on a real sorted WordPress tree.

## Benchmarks

Direct in-process `file_fetch` microbenchmark against `packages/reprint-exporter/src/export.php`, PHP 8.2.29, output redirected to `/dev/null`.

| Dataset | Trunk | This branch | Delta | Notes |
|---|---:|---:|---:|---|
| Favorable mixed batch: 32 MiB random `.jpg` + repetitive `.css` | 930ms | 85ms | 10.9x faster | branch avoids gzipping random binary bytes |
| Unfavorable binary-only batch: 32 MiB random `.jpg` | 94ms | 78ms | roughly neutral | both stay identity |

I also tested importer-side class splitting on the current `large-directory` e2e fixture. It was rejected: trunk completed `files-pull` in 49.472s with 1 fetch request, while the naive split created 125 sequential fetch requests and took 65.753s.

## Validation

- `php -l packages/reprint-exporter/src/export.php`
- `php -l tests/FileFetchCompressionTest.php`
- `vendor/bin/phpunit tests/FileFetchCompressionTest.php --colors=never`

### DIFF
--- a/packages/reprint-exporter/src/export.php
+++ b/packages/reprint-exporter/src/export.php
@@ -3396,23 +3396,18 @@ function endpoint_file_fetch(
  *     input). Negligible per individual file, but unbounded if the batch is
  *     all-binary multiplied by request volume.
  *
- * Rule: gzip the response if **any** file in the batch is compressible.
+ * Rule: gzip the response only if every file in the batch is compressible.
  *
- * The previous all-or-nothing rule ("gzip only if every file is compressible")
- * was over-conservative — a single PNG in a 200-CSS batch flipped the whole
- * response to identity, losing ~50 % of wire size that would have compressed.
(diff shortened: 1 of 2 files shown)


adamziel:
Closing for this db-apply speedup pass. This file_fetch compression idea is backed by an in-process file_fetch microbenchmark and does not target the accepted full database apply gate. The surviving db-apply speedups are #235 and #239.

PR #214, opened 2026-05-11
[codex] Add per-part gzip for mixed file fetches

adamziel:
## What it does

Adds an opt-in `file_part_gzip` capability for `file_fetch` mixed batches. When the importer sends the capability, the exporter keeps the HTTP response identity and gzip-encodes only compressible file-part bodies with `X-Body-Encoding: gzip`.

Text-only batches still use response-level gzip. Binary-only batches stay identity. Mixed batches without the capability keep the old response-level gzip behavior for older importers.

## Rationale

Response-level gzip is all-or-nothing. In mixed media/text batches it preserves wire size, but it burns exporter CPU deflating already-compressed media bytes. Per-part gzip keeps one sequential request while avoiding that binary CPU cost.

Benchmark, PHP 8.2, gzip level 6, median of 5 synthetic runs:

| Dataset | Old response gzip encode+decode | Hybrid encode+decode | Wire size |
| --- | ---: | ---: | --- |
| 4 MiB text + 28 MiB binary | 1311.6 ms | 409.5 ms | 87.55% vs 87.57% of identity |
| 28 MiB text + 4 MiB binary | 527.7 ms | 299.9 ms | 12.83% vs 12.81% of identity |
| 2000 small text files | 52.6 ms | 52.6 ms | unchanged; still response gzip |
| 32 MiB binary | identity | identity | unchanged |

## Implementation

- Adds `file_fetch_compression_mode()` with three modes: `identity`, `response-gzip`, and `file-parts`.
- Sends `file_part_gzip=1` from the importer for `file_fetch` requests.
- Parses `file_part_gzip` as a boolean in the exporter HTTP server.
- Buffers gzip-encoded file parts in the importer, decodes them, then emits the same streaming body + streaming close events as raw file parts so checkpoint behavior stays aligned.
- Leaves old importers compatible: no capability means mixed batches still use response-level gzip.

## Testing instructions

Passed:

```bash
php -l packages/reprint-exporter/src/class-http-server.php
php -l packages/reprint-exporter/src/export.php
php -l tests/FileFetchCompressionTest.php
php -l tests/Import/FileBodyStreamingTest.php
vendor/bin/phpunit -c tests/phpunit.xml tests/FileFetchCompressionTest.php tests/Import/FileBodyStreamingTest.php tests/Import/MultipartStreamParserTest.php tests/Import/CurlTimeoutRecoveryTest.php tests/ExportHttpServerTest.php
php -r 'require "packages/reprint-exporter/src/class-http-server.php"; $s=new Site_Export_HTTP_Server(); $config=$s->parse_http_config(["file-part-gzip"=>"1"],[],[],""); if (($config["file_part_gzip"] ?? null) !== true) exit(1);'
composer run-script analyze
```

Also ran full `vendor/bin/phpunit -c tests/phpunit.xml`; it still fails in this environment because MySQL dump tests cannot authenticate as local `root`. `composer run-script compat` still reports existing PHPCompatibility findings under vendored `data-liberation/vendor-patched` code.

### DIFF
--- a/packages/reprint-exporter/src/class-http-server.php
+++ b/packages/reprint-exporter/src/class-http-server.php
@@ -171,7 +171,7 @@ final class Site_Export_HTTP_Server {
                 $value = (int) $value;
             } elseif (in_array($key, ['memory_threshold'], true)) {
                 $value = (float) $value;
-            } elseif (in_array($key, ['create_table_query', 'db_unbuffered', 'follow_symlinks'], true)) {
+            } elseif (in_array($key, ['create_table_query', 'db_unbuffered', 'follow_symlinks', 'file_part_gzip'], true)) {
                 $value = filter_var($value, FILTER_VALIDATE_BOOLEAN);
             } elseif ($key === 'paths' && is_string($value)) {
                 $decoded = json_decode($value, true);
--- a/packages/reprint-exporter/src/export.php
+++ b/packages/reprint-exporter/src/export.php
@@ -3399,7 +3399,11 @@ function endpoint_file_fetch(
         $producer,
         $budget,
         $config,
-        file_fetch_compression_mode($paths, !empty($config["file_part_gzip"])),
+        file_fetch_compression_mode(
+            $paths,
+            !empty($config["file_part_gzip"]),
+            $directories
+        ),
     );
 }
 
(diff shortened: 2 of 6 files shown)


adamziel:
HTTP wrapper benchmark follow-up after `dcc13de`.

I added `tests/e2e/benchmark/bench-file-fetch-wrappers.mjs`, which starts the real `file_fetch` endpoint behind PHP's built-in server and local Node reverse-proxy wrappers. Each scenario runs 1 warmup + 5 measured sequential requests and verifies the multipart payload byte-for-byte.

Key patched results, new mode vs old mode:

| Wrapper | Balanced mixed | Binary-heavy mixed | Text-heavy many-small mixed |
|---|---:|---:|---:|
| PHP default | -73.4% | -86.1% | +1.1% |
| PHP zlib/output buffering on | -73.1% | -86.7% | +1.5% |
| Proxy pass-through | -74.1% | -81.8% | -11.3% |
| Proxy buffering | -70.0% | -81.3% | -13.2% |
| Proxy gzips identity responses | +24.2% | +9.7% | +0.4% |

The original version was bad for `2000` tiny text files + one tiny binary: up to +81.9% wall time and +1468% wire bytes because per-part gzip created thousands of independent gzip members. The new size/count heuristic keeps that shape on response gzip, while preserving the large binary-heavy wins.

Breakpoints found:

- A wrapper that strips HTTP `Content-Encoding` breaks response-gzip for both old and new clients.
- A body-inspecting wrapper that strips `X-Body-Encoding` breaks per-part gzip. The importer now catches that with a final file-size mismatch instead of silently indexing compressed bytes.
- With a 24 MiB text chunk + 1 MiB binary and `chunk_size=32 MiB`, PHP `memory_limit` at 32/48/64 MiB fails in both old and new modes; 96 MiB passes. This is not specific to per-part gzip, but it defines the low-memory/chunk-size cliff.

Verification run:

- `vendor/bin/phpunit -c tests/phpunit.xml tests/FileFetchCompressionTest.php tests/Import/FileBodyStreamingTest.php tests/Import/MultipartStreamParserTest.php tests/Import/CurlTimeoutRecoveryTest.php tests/ExportHttpServerTest.php`
- `composer run-script analyze`
- `node --check tests/e2e/benchmark/bench-file-fetch-wrappers.mjs`
- `BENCH_REPS=5 BENCH_OUT=/tmp/reprint-file-fetch-wrappers-after.json node tests/e2e/benchmark/bench-file-fetch-wrappers.mjs`

adamziel:
Closing for this db-apply speedup pass. This is a file_fetch compression experiment, not a database apply phase improvement. The surviving db-apply speedups are #235 and #239, both verified with full playground-sqlite-db-apply timings.

PR #216, opened 2026-05-18
Measure native batch URL rewrite impact

adamziel:
## What it does

Uses the optional `wp_native_apis_rewrite_text_url_bases()` extension function from WordPress/php-toolkit#282 when Reprint rewrites plain-text URL values.

Also changes the PR performance workflow to compare the same Reprint artifact with and without the native extension loaded on both host PHP and PHP.wasm.

## Rationale

The previous native URL path still made PHP drive URL detection and rewriting one URL at a time. This PR measures a coarser boundary: pass the whole decoded value and whole URL mapping into the native extension, then get the rewritten value back in one call.

## Implementation

`StructuredDataUrlRewriter` builds a compact mapping string once in the constructor. For `plain_text`, and for `block_markup` values that are actually text-only, it calls the native batch function when the `wp_native_apis` extension is loaded. If the extension is unavailable, the existing PHP path is used unchanged.

The performance workflow now:

- builds `wp_native_apis` from `WordPress/php-toolkit` branch `codex-native-structured-rewrite`
- benchmarks host PHP userland vs host PHP native extension
- builds and serves a local PHP.wasm extension manifest
- benchmarks PHP.wasm userland vs PHP.wasm native extension
- posts one sticky PR comment with the deltas

## Benchmarks

Latest CI run: https://github.com/adamziel/reprint/actions/runs/26067523070/attempts/3

Focused case: `StructuredDataUrlRewriter` rewrites 10,000 URLs across 556,690 scanned bytes.

| Runtime | Userland | Native extension | Delta |
|---|---:|---:|---:|
| Host PHP | 2968.918 ms | 6.743 ms | -99.8%, about 440x faster |
| PHP.wasm 8.4 | 7616.311 ms | 7.632 ms | -99.9%, about 998x faster |

These are single-run CI measurements, so treat them as directional. They do show the intended effect: PHP no longer iterates URL-by-URL for plain text URL base rewrites.

## Testing instructions

Local checks run:

```bash
vendor/bin/phpunit tests/UrlRewriting/StructuredDataUrlRewriterTest.php
php tests/e2e/benchmark/structured-text-url-rewrite-bench.php
vendor/bin/phpstan analyze --memory-limit=1G
node --check tests/e2e/benchmark/bench-pull.mjs
node --check tests/e2e/benchmark/render-diff.mjs
node --check tests/e2e/ci/php-wasm-runner.mjs
git diff --check
composer build:phar
```

Local fallback benchmark result without the extension:

```json
{"total_urls_rewritten":10000,"elapsed_ms":3233.079262,"native_batch_text_rewrite":"no"}
```

CI checks are green on the Reprint PR, including PHPUnit, E2E, PHPStan, and the host/PHP.wasm performance workflow. I could not measure the native extension locally because this environment has no `php-config` and no package manager.

Depends on WordPress/php-toolkit#282.

### DIFF
--- a/.github/workflows/perf-report.yml
+++ b/.github/workflows/perf-report.yml
@@ -7,9 +7,15 @@ permissions:
   contents: read
   pull-requests: write
 
+env:
+  PHP_VERSION: '8.3'
+  WASM_PHP_VERSION: '8.4'
+  PHP_TOOLKIT_NATIVE_REF: codex-native-structured-rewrite
+  WP_MYSQL_PARSER_EXTENSION_MANIFEST_URL: https://wordpress.github.io/sqlite-database-integration/wp_mysql_parser-wasm-extension/latest/manifest.json
+
 jobs:
   build-pr-phar:
-    name: Build reprint.phar (PR)
+    name: Build reprint.phar
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v5
@@ -18,7 +24,7 @@ jobs:
 
       - uses: shivammathur/setup-php@v2
         with:
-          php-version: '8.5'
+          php-version: ${{ env.PHP_VERSION }}
           extensions: pdo, pdo_mysql, json, hash, zlib, mbstring
 
       - run: composer install --no-dev --no-interaction --prefer-dist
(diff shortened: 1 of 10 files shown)


adamziel:
Closing under the current rule for this thread. This PR is centered on isolated/native URL rewrite benchmarks and is conflict-dirty; it does not provide surviving evidence on the accepted full playground-sqlite-db-apply gate. Current db-apply survivors are #235 and #239.

PR #217, opened 2026-05-19
Cache structured URL rewrite results

adamziel:
## What

Adds a bounded 4096-entry ring cache to `StructuredDataUrlRewriter` for URL rewrite decisions.

For each matched raw URL, the importer now caches whether it maps to a target URL and, when it does, the rewritten raw URL plus parsed target URL. The cache key includes the rewriter's URL mapping, content type, token type, and raw URL so block markup and plain text keep their existing semantics.

## Why

Large imports often repeat the same source URLs many times. Without a cache, every occurrence repeats `is_child_url_of()` and `WPURL::replace_base_url()` work even when the result is identical.

This complements WordPress/php-toolkit#283, which makes `URLInTextProcessor` use the native URL candidate scanner and adds a parsed-candidate cache in the shared DataLiberation layer.

## Validation

- `php -l packages/reprint-importer/src/lib/url-rewrite/class-structured-data-url-rewriter.php`
- `vendor/bin/phpcs --standard=PHPCompatibility --runtime-set testVersion 7.4- -ps packages/reprint-importer/src/lib/url-rewrite/class-structured-data-url-rewriter.php`
- `../vendor/bin/phpunit --testdox --colors=never UrlRewriting/StructuredDataUrlRewriterTest.php`

The focused URL rewriter suite passed: 39 tests, 48 assertions, 7 skipped.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -8,7 +8,7 @@
     "packages": [
         {
             "name": "wp-php-toolkit/bytestream",
-            "version": "v0.7.9",
+            "version": "v0.8.0",
             "source": {
                 "type": "git",
                 "url": "https://github.com/wp-php-toolkit/bytestream.git",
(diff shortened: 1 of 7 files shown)


adamziel:
Updated this branch onto the new `wp-php-toolkit/*` `v0.8.0` release and re-ran the focused large DB apply benchmark.

Environment:
- `large-directory` Docker e2e site
- 320,007 posts, 720,015 postmeta rows, 2,000+ files
- PHP.wasm 8.4 for `playground-sqlite-db-apply`
- `wp_mysql_parser` native extension loaded from the sqlite-database-integration WASM manifest
- `wp_native_apis` native extension loaded from the php-toolkit WASM manifest

| Stage | Wall time | Attempts | Status | Notes |
|---|---:|---:|---|---|
| `playground-sqlite-db-apply` | 335.22 s | 1 | pass | native MySQL lexer/parser verified; SQLite driver parser verified |

I also added the benchmark runner hook for `WP_NATIVE_APIS_EXTENSION_MANIFEST` and mounted the repo root in PHP.wasm so the sqlite-database-integration submodule is visible during the WASM run.

adamziel:
Closing as most of the benefit was gained through other two PRs

PR #220, opened 2026-05-20
Speed up SQLite db-apply imports

adamziel:
## What it does

Speeds up SQLite `db-apply` imports while keeping URL rewrites structured:

- SQLite no longer executes PHP-backed `FROM_BASE64()` for every imported string value. For SQLite targets, scanner-identified `FROM_BASE64()` / `CONVERT(FROM_BASE64() USING utf8mb4)` expressions are rewritten to MySQL-compatible `0x...` hex literals before execution.
- URL rewriting no longer uses byte-level literal source-domain replacement. Plain-text values still go through the URL scanner, WHATWG URL parsing, `is_child_url_of()`, and `WPURL::replace_base_url()`.
- Reprint does not infer Markdown, CSS, CSV, TOML, or other container syntax in the plain-text layer. Known structured formats are parsed by their owning parser first; only leaf text or intentionally unknown text is delegated to `URLInTextProcessor`.
- Source-domain quick rejection considers both literal hosts and WPURL-normalized hosts, so Unicode source mappings still match punycode URLs in the dump.

## Rationale

The first high-speed URL path used literal base URL replacement after boundary checks. That recovered a lot of speed, but it was not semantically equivalent to structured URL rewriting: IDN/punycode and other URL-normalization cases can be missed or rewritten incorrectly. This PR removes that path and keeps URL decisions centralized in structured URL parsing.

A later experiment split delimiter-looking suffixes from `URLInTextProcessor` candidates. That also has been removed. It was not rigorous because reprint would have been guessing syntax that belongs to typed parsers. The invariant now is that `URLInTextProcessor` owns plain-text URL span detection; reprint callers do not repair or reinterpret those spans.

For SQLite imports, the safe big win remains avoiding per-value PHP user-defined function calls for `FROM_BASE64()`. `Base64ValueScanner` already has SQL-tokenized expression boundaries, so it can replace the full transport expression with a hex literal without regexp SQL rewriting or SQL string escaping.

## Implementation

`Base64ValueScanner` records both the quoted payload range and the full outer expression range. After URL rewriting, SQLite imports can ask the scanner to render decoded values as hex literals:

```sql
CONVERT(FROM_BASE64('Qm9iJ3MgVGVzdCBCbG9n') USING utf8mb4)
```

becomes:

```sql
0x426f622773205465737420426c6f67
```

`StructuredDataUrlRewriter` handles structured layers in order: PHP serialization, JSON, SQL/base64-decoded leaf values, WordPress block markup when hinted or auto-detected, and finally plain text through `URLInTextProcessor`. It keeps bounded rewrite-result caches, but it does not do raw string URL replacement.

## Benchmarks

PHP.wasm 8.4 db-apply runs on the 262 MB SQL fixture with both native manifests loaded (`wp-native-apis-playground-extension` and `wp_mysql_parser-wasm-extension`):

| Run | Native manifests | `db-apply` wall time | Notes |
|---|---:|---:|---|
| No URL rewrite control | Yes | 31.61 s | SQLite execution + base64 hex inlining |
| Current rigorous PR | Yes | 106.93 s | No literal replacement, no delimiter guessing |
| Previous native structured-only build | Yes | 113.34 s | Before punycode quick-reject fix |
| Current rigorous PR | No | 852.95 s | Same PHAR without native manifests |

Historical local source replay on the same fixture, before the literal URL fast path was removed:

| Run | `db-apply` wall time | Delta |
|---|---:|---:|
| Baseline `v0.8.0` source replay | 121.37 s | - |
| Hex inlining + literal URL fast path | 40.12 s | -81.25 s (-67.0%) |
| Built PHAR after scanner hardening | 40.38 s | -80.99 s (-66.7%) |
| After removing literal URL replacement fast path, without native URL scanner | 380.41 s | +259.04 s (+213.4%) |

The native benchmark is the relevant Studio direction: with the native URL scanner and native MySQL parser loaded, structured-only rewriting avoids the unsafe literal URL replacement path while keeping the run much faster than userland-only structured URL rewriting.

## Hot paths

A rewrite-only profiler over the same 262 MB SQL dump, without SQLite execution, attributed the remaining cost as follows:

| Area | Count | Time |
|---|---:|---:|
| Structured plain-text leaf rewrites | 201,505 values | 56.98 s |
| Structured block-markup rewrites | 49,753 values | 17.46 s |
| Fast INSERT scan | 824 fast statements | 2.93 s |
| Base64 decode for URL candidates | 251,258 values | 0.52 s |
| Source-domain prefilter | 251,258 values | 0.12 s |
| Hex literal rendering | 824 fast statements | 0.63 s |

`scanner_walk` in the profiler is inclusive of the structured rewrite calls; the main actionable sink is the 201k plain-text leaf rewrites, followed by block markup.

A post-content microbench also confirms the bounded rewrite-result cache matters: 5,000 block values / 350,000 URL occurrences / 2 unique raw URLs take 2.03 s with caching versus 15.91 s without. An exact whole-leaf URL fast path was tested and rejected because it made the rewrite-only profile worse (99.79 s → 106.28 s).

## Proposed next optimization

The next rigorous speedup should avoid guessing syntax and instead reduce work before plain-text URL scanning:

- Add explicit SQL column content hints for columns that are known leaf text, not containers. Those can skip failed serialized-PHP and JSON probes and go directly to `URLInTextProcessor`. Unknown columns and `postmeta.meta_value` should keep auto-detection.
- Move more of the plain-text rewrite loop into php-toolkit/native only if the native API owns the same span contract: scan validated URL spans, parse, child-check, replace, and cache in one place. That would cut PHP/native crossings and PHP object churn without moving Markdown/CSS/CSV guessing into reprint.
- Keep the current bounded caches and measure a content-level leaf rewrite cache separately. It is semantically safe because it keys by exact input value and content type, but it is only worth adding if the dump has enough repeated decoded values to beat hashing overhead.

Fixture details:

- WordPress `6.8.2-php8.3-apache` + MariaDB 11 source site.
- 52,003 posts and 154,002 postmeta rows.
- 2,000 attachment database records.
- SQL dump: 262 MB. Resulting SQLite DB: 223 MB.

Validation on the generated SQLite DB:

```sql
pragma integrity_check; -- ok
select count(*) from wp_posts; -- 52003
select count(*) from wp_postmeta; -- 154002
```

Checked URL-heavy columns after import: no remaining `127.0.0.1:8199` in `wp_posts.guid`, `wp_posts.post_content`, or `wp_postmeta.meta_value`.

## Testing instructions

```bash
composer test -- --filter StructuredDataUrlRewriterTest
composer test -- --filter 'StructuredDataUrlRewriterTest|SqlStatementRewriterTest|SqlStatementRewriterPrefilterTest|SqlStatementRewriterLexerWalkerTest|NewSiteUrlSqliteTest'
composer analyze
composer build:phar
```

The broader filtered suite currently exits successfully but still reports the existing `NewSiteUrlSqlite` output warnings.

### DIFF
--- a/packages/reprint-importer/src/import.php
+++ b/packages/reprint-importer/src/import.php
@@ -4990,7 +4990,7 @@ class ImportClient
 
     private function inline_base64_values_for_sqlite(string $query): string
     {
-        if (strpos($query, "FROM_BASE64(") === false) {
+        if (stripos($query, "FROM_BASE64") === false) {
             return $query;
         }
 
--- a/packages/reprint-importer/src/import.php
+++ b/packages/reprint-importer/src/import.php
@@ -4995,7 +4995,7 @@ class ImportClient
         }
 
         $scanner = new Base64ValueScanner($query);
-        return $scanner->get_result_with_mysql_hex_literals();
+        return $scanner->get_result_with_sqlite_compatible_literals();
     }
 
     public function run_db_apply(array $options): void
--- a/packages/reprint-importer/src/import.php
+++ b/packages/reprint-importer/src/import.php
@@ -4988,24 +4988,14 @@ class ImportClient
         ];
     }
 
-    private function inline_from_base64_for_sqlite(string $query): string
+    private function inline_base64_values_for_sqlite(string $query): string
     {
         if (strpos($query, "FROM_BASE64(") === false) {
             return $query;
         }
 
-        return preg_replace_callback(
-            "/(?:CONVERT\\s*\\(\\s*)?FROM_BASE64\\s*\\(\\s*'([A-Za-z0-9+\\/=]*)'\\s*\\)(?:\\s+USING\\s+utf8mb4\\s*\\))?/i",
-            static function (array $matches): string {
-                $decoded = base64_decode($matches[1], true);
-                if ($decoded === false) {
-                    return $matches[0];
-                }
-
-                return "'" . str_replace("'", "''", $decoded) . "'";
-            },
-            $query
-        );
+        $scanner = new Base64ValueScanner($query);
+        return $scanner->get_result_with_mysql_hex_literals();
     }
 
     public function run_db_apply(array $options): void
@@ -5138,7 +5128,7 @@ class ImportClient
         }
 
         [$pdo, $connection_label] = $this->create_target_db_apply_connection($options);
-        $inline_from_base64_for_sqlite = ($options["target_engine"] ?? "mysql") === "sqlite";
+        $inline_base64_values_for_sqlite = ($options["target_engine"] ?? "mysql") === "sqlite";
 
         $this->audit_log(
             "CONNECTED | {$connection_label}",
@@ -5245,10 +5235,9 @@ class ImportClient
 
                     // Rewrite URLs if mapping is configured
                     if ($stmt_rewriter) {
-                        $query = $stmt_rewriter->rewrite($query);
-                    }
-                    if ($inline_from_base64_for_sqlite) {
-                        $query = $this->inline_from_base64_for_sqlite($query);
+                        $query = $stmt_rewriter->rewrite($query, $inline_base64_values_for_sqlite);
+                    } elseif ($inline_base64_values_for_sqlite) {
+                        $query = $this->inline_base64_values_for_sqlite($query);
                     }
 
                     // Execute against target database
@@ -5324,10 +5313,9 @@ class ImportClient
                 }
 
                 if ($stmt_rewriter) {
-                    $query = $stmt_rewriter->rewrite($query);
-                }
-                if ($inline_from_base64_for_sqlite) {
-                    $query = $this->inline_from_base64_for_sqlite($query);
+                    $query = $stmt_rewriter->rewrite($query, $inline_base64_values_for_sqlite);
+                } elseif ($inline_base64_values_for_sqlite) {
+                    $query = $this->inline_base64_values_for_sqlite($query);
                 }
 
                 try {
--- a/packages/reprint-importer/src/import.php
+++ b/packages/reprint-importer/src/import.php
@@ -4988,6 +4988,26 @@ class ImportClient
         ];
     }
 
+    private function inline_from_base64_for_sqlite(string $query): string
+    {
+        if (strpos($query, "FROM_BASE64(") === false) {
+            return $query;
+        }
+
+        return preg_replace_callback(
+            "/(?:CONVERT\\s*\\(\\s*)?FROM_BASE64\\s*\\(\\s*'([A-Za-z0-9+\\/=]*)'\\s*\\)(?:\\s+USING\\s+utf8mb4\\s*\\))?/i",
+            static function (array $matches): string {
+                $decoded = base64_decode($matches[1], true);
+                if ($decoded === false) {
+                    return $matches[0];
+                }
+
+                return "'" . str_replace("'", "''", $decoded) . "'";
+            },
+            $query
+        );
+    }
+
     public function run_db_apply(array $options): void
     {
         $sql_file = $this->state_dir . "/db.sql";
@@ -5118,6 +5138,7 @@ class ImportClient
         }
 
         [$pdo, $connection_label] = $this->create_target_db_apply_connection($options);
+        $inline_from_base64_for_sqlite = ($options["target_engine"] ?? "mysql") === "sqlite";
 
         $this->audit_log(
             "CONNECTED | {$connection_label}",
@@ -5226,6 +5247,9 @@ class ImportClient
                     if ($stmt_rewriter) {
                         $query = $stmt_rewriter->rewrite($query);
                     }
+                    if ($inline_from_base64_for_sqlite) {
+                        $query = $this->inline_from_base64_for_sqlite($query);
+                    }
 
                     // Execute against target database
                     try {
@@ -5301,6 +5325,9 @@ class ImportClient
 
                 if ($stmt_rewriter) {
                     $query = $stmt_rewriter->rewrite($query);
+                }
+                if ($inline_from_base64_for_sqlite) {
+                    $query = $this->inline_from_base64_for_sqlite($query);
                 }
 
                 try {
--- a/packages/reprint-importer/src/lib/url-rewrite/class-base64-value-scanner.php
+++ b/packages/reprint-importer/src/lib/url-rewrite/class-base64-value-scanner.php
@@ -8,8 +8,9 @@
  * Detects CONVERT(FROM_BASE64('...') USING utf8mb4) wrappers automatically.
  * set_value() only replaces the base64 payload inside the quotes, so wrappers
  * are preserved without the caller needing to know about them. SQLite imports
- * may instead replace the entire outer expression with a MySQL-compatible hex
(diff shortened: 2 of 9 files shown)


adamziel:
Added the extra plain-text data-format coverage and pushed it in `5a369e4`.

New cases cover Markdown links, CSS `url(...)`, JS object literals, WordPress shortcode attributes, CSV fields, INI assignments, YAML sequence items, XML text nodes, TOML quoted values, and CDATA. I also added an IDN/punycode source-domain case because that is the kind of URL-normalization bug the literal replacement path would miss.

The fix keeps the native/PHP URL scanner path, but applies replacements by original byte offsets instead of asking `URLInTextProcessor` to replace the whole scanner span. When the scanner exposes plain-text delimiters as part of a broad candidate, we split those bytes from the URL candidate, run the structured parse/child/rewrite path on the URL prefix, then append the delimiters unchanged. No `str_replace`/`preg_replace` URL rewrite path is reintroduced.

Verification:

| Check | Result |
|---|---:|
| `composer test -- --filter StructuredDataUrlRewriterTest` | 55 tests / 70 assertions, 7 skipped |
| SQL/db-apply filtered suite | 140 tests / 2130 assertions, exits 0; existing `NewSiteUrlSqlite` output warnings remain |
| `composer analyze` | no errors |
| `composer build:phar` | passed |

Latest native PHP.wasm benchmark on the 262 MB fixture with both native manifests loaded:

| Run | Wall time |
|---|---:|
| Previous native build before this hardening | 113.34 s |
| Current PR after delimiter + punycode fixes | 110.20 s |

adamziel:
Updated the PR to match the stricter ownership model.

Changes pushed in `6a998ac`:

- Removed the reprint-side `split_plain_text_url_candidate()` delimiter heuristic.
- Removed the Markdown/CSS/CSV/TOML-style tests because those implied reprint was parsing typed container syntax in the plain-text layer.
- Kept the punycode/IDN source-domain test, since that belongs to reprint's source-domain prefilter contract.
- Updated an outdated SQL rewriter comment so it no longer mentions a literal URL fast path.

I also documented the rigor rule in the Studio workspace `AGENTS.md` on https://github.com/Automattic/studio/pull/3248: known structured formats must be parsed by their owning parser, and only leaf/unknown text may be delegated to `URLInTextProcessor`. No `str_replace`, `preg_replace`, delimiter trimming, or format guessing for URL rewrites.

Latest measurements on the 262 MB SQL fixture with both native manifests loaded:

| Run | Wall time |
|---|---:|
| No URL rewrite control | 31.61 s |
| Current rigorous PR | 106.93 s |
| Previous native structured-only build | 113.34 s |
| Current rigorous PR without native manifests | 852.95 s |

Rewrite-only profile, excluding SQLite execution:

| Area | Count | Time |
|---|---:|---:|
| Structured plain-text leaf rewrites | 201,505 values | 56.98 s |
| Structured block-markup rewrites | 49,753 values | 17.46 s |
| Fast INSERT scan | 824 statements | 2.93 s |
| Base64 decode for URL candidates | 251,258 values | 0.52 s |
| Source-domain prefilter | 251,258 values | 0.12 s |
| Hex literal rendering | 824 statements | 0.63 s |

I tested one rigorous optimization candidate, an exact whole-leaf URL fast path, and rejected it because it made the rewrite-only profile worse (`99.79 s` -> `106.28 s`). The next promising direction is explicit SQL column content hints for columns known to be leaf text, so those skip failed serialized/JSON probes and go straight to `URLInTextProcessor`; unknown columns and `postmeta.meta_value` should keep auto-detection.

adamziel:
Speedup experiments stacked on this PR:

| PR | Idea | Full `db-apply` | Rewrite-only profile | Result |
| --- | --- | ---: | ---: | --- |
| #221 | Skip block-markup parser when a `block_markup` value contains no `<` byte, then treat it as leaf text | 94.24 s | 90.15 s | best result, -11.9% full run vs this PR |
| #222 | Use `WHOLE_URL` hints for schema-known URL columns (`guid`, comment author URL, links URL fields) | 103.73 s | 99.37 s | small full-run win, rewrite-only flat |
| #223 | Gate serialized-PHP/JSON parser probes by required starter bytes | not run | 101.03 s | measured rejection, rewrite-only slower |

Baseline for comparison is this PR at 106.93 s full `db-apply` and 99.79 s rewrite-only on the 262 MB SQL fixture with PHP.wasm 8.4 and both native manifests loaded.

The only experiment with a meaningful win so far is #221. It stays within the rigor contract: known block/HTML syntax still belongs to `BlockMarkupUrlProcessor`; values with no `<` byte have no possible HTML tag or WordPress block-comment opener, so the remaining content is leaf text for `URLInTextProcessor`.

adamziel:
Follow-up exploration: five more isolated ideas measured against this PR's rewrite-only baseline (`99.79s`) on the same 262 MB fixture with PHP.wasm 8.4 and both native manifests loaded.

| Idea | Branch | Rewrite-only | Result |
| --- | --- | ---: | --- |
| Exact absolute URL leaf fast path | `speedup-exact-url-leaves` | 106.00 s | rejected; direct whole-value parse is slower than native-backed text scanning |
| Origin-indexed URL mappings | `speedup-origin-indexed-mapping` | 96.49 s | candidate; avoids checking impossible mappings |
| Carry column names on scanner entries | `speedup-scanner-column-carry` | 113.47 s | rejected; larger entry arrays cost more than binary search saves |
| Map only `FROM_BASE64` values in fast inserts | `speedup-map-only-base64-values` | 95.78 s / 97.52 s rerun | small candidate; noisy but preserves counts |
| Cache scanner entry count | `speedup-scanner-entry-count` | 101.24 s | rejected; not worth it under WASM |

The actionable stack is #224 on top of #221:

| Run | Full `db-apply` | Rewrite-only |
| --- | ---: | ---: |
| This PR | 106.93 s | 99.79 s |
| #221 | 94.24 s | 90.15 s |
| #224 | 89.29 s | 77.06 s |

#224 passes the focused URL rewrite suite plus `composer analyze`; the suite still has the existing `NewSiteUrlSqlite` warning output.

adamziel:
Two larger follow-up ideas explored on top of #224:

| Idea | Rewrite-only profile | Result |
| --- | ---: | --- |
| Guarded literal-origin fast path for plain text leaves | 17.30 s | success; 4.45x vs #224, 5.8x vs #220 |
| Try the same guarded literal leaf rewrite before SQL column-hint resolution | 7.66 s | success; 10.1x vs #224, 13.0x vs #220 |

Draft PR: #225.

Full `db-apply` on #225 is 33.52 s. The no-URL-rewrite control on this fixture is 31.61 s, so URL rewriting is now close to the full-run floor. Further full-run speedups likely need to target SQLite execution/import overhead rather than URL rewriting.

Rejected variants:

| Variant | Rewrite-only profile | Why rejected |
| --- | ---: | --- |
| Index encoded-http base64 candidate entries | 17.80 s | candidate index memory/build cost exceeded loop savings |
| Run literal leaf fast path before parser probes for markup-free block hints | 18.00 s | extra guard work exceeded the parser-probe savings |

adamziel:
Closing the old db-apply stack root. It is conflict-dirty and superseded by the current verified path: #235 at 85.41s and #235 + #239 at 56.86s on the accepted playground-sqlite-db-apply gate with native parser verified. Keeping this old stack open would preserve stale alternatives rather than the surviving full-phase wins.

PR #221, opened 2026-05-20
Use plain text rewriting for markup-free block values

adamziel:
## What it does

For values already classified as `block_markup`, this skips `BlockMarkupUrlProcessor` when the value has no `<` byte and delegates directly to the plain text URL processor.

## Rationale

Under the current rigor contract, `URLInTextProcessor` should only see leaf text or unknown text after structured formats have had first ownership. For a block-markup-hinted value, the owning block/HTML parser has nothing to parse unless the value contains a `<` byte: there can be no HTML tag and no WordPress block-comment opener. This keeps parser ownership intact while avoiding a costly parser stack on plain post content.

## Implementation

The `BLOCK_MARKUP` path now checks for `<` before constructing `BlockMarkupUrlProcessor`. When absent, it treats the value as leaf text and calls the existing `PLAIN_TEXT` path. A regression test covers the delegation.

## Benchmark

| --- | ---: | ---: | --- |
| This branch | 94.24 s | 90.15 s | `structured_rewrite_block` drops from 17.46 s to 6.86 s |
| Delta | -12.69 s (-11.9%) | -9.64 s (-9.7%) | fastest experiment so far |

## Testing instructions

```bash
composer test -- --filter 'StructuredDataUrlRewriterTest|SqlStatementRewriterTest|SqlStatementRewriterPrefilterTest|SqlStatementRewriterLexerWalkerTest|NewSiteUrlSqliteTest'
composer analyze
```

### DIFF
--- a/packages/reprint-importer/src/lib/url-rewrite/class-structured-data-url-rewriter.php
+++ b/packages/reprint-importer/src/lib/url-rewrite/class-structured-data-url-rewriter.php
@@ -318,6 +318,14 @@ class StructuredDataUrlRewriter
 
         switch ( $content_type ) {
             case self::BLOCK_MARKUP:
+                // Without a `<` byte there can be no HTML tag or WordPress
+                // block-comment opener for BlockMarkupUrlProcessor to own.
+                // Treat the value as the leaf text it is instead of running
+                // the block-markup parser stack.
+                if ( strpos( $content, '<' ) === false ) {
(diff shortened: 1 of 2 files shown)


adamziel:
Closing as superseded by the current verified db-apply stack. The recorded full db-apply result here was 94.24s on the old stack, while current #235 is 85.41s and #235 + #239 is 56.86s on the accepted playground-sqlite-db-apply gate with native parser verified.

PR #222, opened 2026-05-20
Add whole URL hints for schema-known URL columns

adamziel:
## What it does

Adds a `WHOLE_URL` content type and uses it for schema-known URL columns such as `wp_posts.guid`, comment author URLs, and links-table URL fields.

## Rationale

These columns store a URL-like value as the full column value, not structured markup. Parsing the entire value as a URL preserves the "URLs are structured data" rule while avoiding the generic text processor for values where the schema already gives us stronger information.

## Implementation

The SQL statement rewriter's column-content map can now mark columns as `WHOLE_URL`. `StructuredDataUrlRewriter` then parses the full value through the structured URL path and rewrites it only when the full value matches the source mapping. Tests cover `wp_posts.guid` rewriting and an embedded source URL that must not be rewritten.

## Benchmark

| --- | ---: | ---: | --- |
| This branch | 103.73 s | 99.37 s | 51,753 values use `WHOLE_URL` |
| Delta | -3.20 s (-3.0%) | -0.42 s (-0.4%) | small full-run win, rewrite-only mostly flat |

## Testing instructions

```bash
php -l packages/reprint-importer/src/lib/url-rewrite/class-sql-statement-rewriter.php
php -l tests/UrlRewriting/SqlStatementRewriterTest.php
composer test -- --filter 'SqlStatementRewriterTest|StructuredDataUrlRewriterTest'
composer analyze
```

### DIFF
--- a/packages/reprint-importer/src/lib/url-rewrite/class-sql-statement-rewriter.php
+++ b/packages/reprint-importer/src/lib/url-rewrite/class-sql-statement-rewriter.php
@@ -19,15 +19,15 @@
  * statements emitted by MySQLDumpProducer follow a constrained set of shapes
  * the walker recognises; for any shape it doesn't, it returns null and every
  * FROM_BASE64() value falls back to plain-text URL rewriting (which is the
- * safe default anyway — block_markup is only ever assigned to a small set of
- * WordPress core columns).
+ * safe default anyway — content-type hints are only assigned to WordPress
+ * core columns with well-defined storage semantics).
  */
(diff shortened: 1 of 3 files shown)


adamziel:
Closing under the current db-apply rule. The recorded full db-apply delta is small (-3.20s / -3.0%) on an old stack and the rewrite-only profile is essentially flat (-0.4%). This does not make enough of a dent to keep as an active speedup PR.

PR #223, opened 2026-05-20
Gate structured parser probes by required starters

adamziel:
## What it does

Adds necessary first-byte gates before trying the serialized-PHP and JSON container parsers in the structured URL rewriter.

## Rationale

This is an experiment to see whether avoiding parser construction for impossible values improves the large `db-apply` path. The checks are necessary conditions only: real parsing and validation still belongs to `PhpSerializationProcessor` and `JsonStringIterator`.

The measurement says this is not a useful speedup on the current fixture, but keeping it as a draft PR makes the rejected idea reviewable and keeps the benchmark record attached to the code.

## Implementation

`can_start_serialized_php()` recognizes only legal PHP serialization starters before the serialized parser is attempted. `can_start_json_container()` skips leading JSON whitespace and requires `{` or `[` because the current JSON iterator accepts container documents.

## Benchmark

| --- | ---: | ---: | --- |
| This branch | not run | 101.03 s | slower in rewrite-only profiling |
| Delta | n/a | +1.24 s (+1.2%) | measured rejection |

## Testing instructions

```bash
composer test -- --filter StructuredDataUrlRewriterTest
composer analyze
```

### DIFF
--- a/packages/reprint-importer/src/lib/url-rewrite/class-structured-data-url-rewriter.php
+++ b/packages/reprint-importer/src/lib/url-rewrite/class-structured-data-url-rewriter.php
@@ -151,32 +151,37 @@ class StructuredDataUrlRewriter
             return $value;
         }
 
-        // Try serialized PHP: the parser validates the entire structure
-        // in the constructor. If it's not malformed, iterate and recurse.
-        $p = new PhpSerializationProcessor($value);
-        if (!$p->is_malformed()) {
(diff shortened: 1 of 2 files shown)


adamziel:
Closing this because the PR body records that the full db-apply measurement is not a useful speedup. Keeping rejected experiments open conflicts with the current rule: only PRs with meaningful full playground-sqlite-db-apply impact should survive.

PR #224, opened 2026-05-20
Index URL mappings and narrow fast column maps

adamziel:
## What it does

Stacks two small URL rewrite hot-loop optimizations on top of #221:

- index parsed URL mappings by origin (`protocol + hostname`) and precompute the decoded parent pathname
- have `FastInsertScanner` add column-map entries only for `FROM_BASE64` expressions, since those are the only offsets the rewriter queries

## Rationale

#221 removes unnecessary block parser work for markup-free block values. After that, the remaining fixture time is dominated by repeated leaf URL rewrite work and scanner bookkeeping. These changes keep the same parser ownership model: URLs are still parsed as structured URLs, and SQL is still walked by the producer-shape scanner or lexer fallback.

## Implementation

`StructuredDataUrlRewriter` now groups mapping entries by parsed origin so each candidate URL only checks mappings that can possibly match. The child-path check still mirrors `is_child_url_of()` but reuses the precomputed parent path.

`FastInsertScanner` now records offset-to-column ranges only for `FROM_BASE64` values. Non-base64 tuple values never reach `Base64ValueScanner`, so their ranges cannot be queried by `SqlStatementRewriter`.

## Benchmark

| --- | ---: | ---: | --- |
| Parent PR #220 | 106.93 s | 99.79 s | rigorous baseline |
| #221 | 94.24 s | 90.15 s | markup-free block values use plain text leaf rewriting |
| This branch | 89.29 s | 77.06 s | #221 plus mapping origin index and narrower fast column map |
| Delta vs #220 | -17.64 s (-16.5%) | -22.73 s (-22.8%) | best stack so far |
| Delta vs #221 | -4.95 s (-5.3%) | -13.09 s (-14.5%) | additive in rewrite-only; smaller but real in full run |

## Testing instructions

```bash
php -l packages/reprint-importer/src/lib/url-rewrite/class-fast-insert-scanner.php
composer test -- --filter 'StructuredDataUrlRewriterTest|SqlStatementRewriterTest|SqlStatementRewriterPrefilterTest|SqlStatementRewriterLexerWalkerTest|Base64ValueScannerTest|NewSiteUrlSqliteTest'
composer analyze
composer build:phar
```

### DIFF
--- a/packages/reprint-importer/src/lib/url-rewrite/class-fast-insert-scanner.php
+++ b/packages/reprint-importer/src/lib/url-rewrite/class-fast-insert-scanner.php
@@ -136,9 +136,9 @@ class FastInsertScanner
                     return null;
                 }
                 $value_end = $cursor;
-                $column_map[] = [$value_start, $value_end, $columns[$col_idx]];
-
                 if (is_array($value_kind)) {
+                    $column_map[] = [$value_start, $value_end, $columns[$col_idx]];
+
                     // FROM_BASE64 payload: kind = [expr_start, expr_length, quote_start, quote_length, encoded_value]
                     $base64_entries[] = [
                         'expr_start' => $value_kind[0],
--- a/packages/reprint-importer/src/lib/url-rewrite/class-structured-data-url-rewriter.php
+++ b/packages/reprint-importer/src/lib/url-rewrite/class-structured-data-url-rewriter.php
@@ -4,7 +4,6 @@ use WordPress\DataLiberation\BlockMarkup\BlockMarkupUrlProcessor;
 use WordPress\DataLiberation\URL\URLInTextProcessor;
 use WordPress\DataLiberation\URL\WPURL;
 
-use function WordPress\DataLiberation\URL\is_child_url_of;
 use function WordPress\DataLiberation\URL\wp_rewrite_urls;
 
 /**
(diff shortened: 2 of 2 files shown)


adamziel:
Two larger follow-up ideas explored on top of #224:

| Idea | Rewrite-only profile | Result |
| --- | ---: | --- |
| Guarded literal-origin fast path for plain text leaves | 17.30 s | success; 4.45x vs #224, 5.8x vs #220 |
| Try the same guarded literal leaf rewrite before SQL column-hint resolution | 7.66 s | success; 10.1x vs #224, 13.0x vs #220 |

Draft PR: #225.

Full `db-apply` on #225 is 33.52 s. The no-URL-rewrite control on this fixture is 31.61 s, so URL rewriting is now close to the full-run floor. Further full-run speedups likely need to target SQLite execution/import overhead rather than URL rewriting.

Rejected variants:

| Variant | Rewrite-only profile | Why rejected |
| --- | ---: | --- |
| Index encoded-http base64 candidate entries | 17.80 s | candidate index memory/build cost exceeded loop savings |
| Run literal leaf fast path before parser probes for markup-free block hints | 18.00 s | extra guard work exceeded the parser-probe savings |

adamziel:
Closing as superseded by the current verified db-apply stack. The recorded full db-apply result here was 89.29s on the old stack, while current #235 is 85.41s and #235 + #239 is 56.86s on the accepted playground-sqlite-db-apply gate with native parser verified.

PR #225, opened 2026-05-20
Fast-path literal plain text URL leaves

adamziel:
## What it does

Adds a guarded literal-origin fast path for unstructured plain text leaves, then calls it from the SQL rewrite loop before column-hint resolution.

This is stacked on #224.

## Rationale

After #224, the remaining rewrite-only profile was dominated by `URLInTextProcessor` dispatch for values that are plain leaves:

- `wp_posts.guid`: whole absolute URLs
- `wp_postmeta.meta_value`: plain prose with a literal source URL
- markup-free `wp_posts.post_content`: leaf text after #221

Those values do not need container parsing or URL candidate scanning when the source origin appears literally at a safe text boundary. The fast path is deliberately narrow and falls back for anything that could be structured syntax.

## Implementation

`StructuredDataUrlRewriter::rewrite_plain_text_literal_leaf()` rewrites only when all of these are true:

- the mapping source is a plain HTTP(S) origin with no source path/query/fragment
- the value has no obvious structured delimiters (`< > " ' \ { } [ ] ( )`)
- the source origin starts at the beginning of the value or after whitespace
- the byte after the source origin is `/`, `?`, `#`, or end-of-string

Everything else returns `false` and continues through the existing parser-owned path.

`SqlStatementRewriter` tries that leaf fast path immediately after decoding and source-domain filtering. If it succeeds, it skips column lookup and full structured dispatch for that value. If it fails, the existing column-aware structured rewrite path runs unchanged.

## Benchmark

| --- | ---: | ---: | --- |
| Parent #220 | 106.93 s | 99.79 s | rigorous baseline |
| #224 | 89.29 s | 77.06 s | previous best stack |
| Literal leaf only | not run | 17.30 s | first explored idea |
| This branch | 33.52 s | 7.66 s | literal leaf before column hints |
| No URL rewrite control | 31.61 s | n/a | practical full-run floor on this fixture |

This clears 10x in the rewrite-only profile versus #224. Full `db-apply` is now close to the no-rewrite floor, so additional full-run gains need to target SQLite execution/import overhead rather than URL rewriting.

Rejected follow-up ideas:

- indexing encoded-http base64 candidates: `17.30 s → 17.80 s` rewrite-only
- pre-parser block hint literal leaf handling: `17.30 s → 18.00 s` rewrite-only

## Testing instructions

```bash
php -l packages/reprint-importer/src/lib/url-rewrite/class-sql-statement-rewriter.php
composer test -- --filter 'StructuredDataUrlRewriterTest|SqlStatementRewriterTest|SqlStatementRewriterPrefilterTest|SqlStatementRewriterLexerWalkerTest|Base64ValueScannerTest|NewSiteUrlSqliteTest'
composer analyze
composer build:phar
```

### DIFF
--- a/packages/reprint-importer/src/lib/url-rewrite/class-sql-statement-rewriter.php
+++ b/packages/reprint-importer/src/lib/url-rewrite/class-sql-statement-rewriter.php
@@ -200,6 +200,14 @@ class SqlStatementRewriter
                     continue;
                 }
 
+                $literal_rewrite = $this->url_rewriter->rewrite_plain_text_literal_leaf($value);
+                if ($literal_rewrite !== false) {
+                    if ($literal_rewrite !== $value) {
+                        $scanner->set_value($literal_rewrite);
+                    }
+                    continue;
+                }
+
                 // Determine content type hint for this column
                 $content_type = null;
                 if ($value_to_column_map !== null) {
--- a/packages/reprint-importer/src/lib/url-rewrite/class-structured-data-url-rewriter.php
+++ b/packages/reprint-importer/src/lib/url-rewrite/class-structured-data-url-rewriter.php
@@ -38,6 +38,15 @@ class StructuredDataUrlRewriter
     private array $source_domains;
 
     /**
+     * Literal origin rewrites that are safe to apply before the generic text
+     * URL scanner. These entries only exist for source mappings whose old base
+     * is an HTTP(S) origin with no path/query/fragment.
+     *
+     * @var list<array{from: string, to: string}>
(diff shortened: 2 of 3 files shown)


adamziel:
Closing under the current accepted full db-apply gate. I reran this branch with BENCH_STAGES=playground-sqlite-db-apply, PHP.wasm 8.3, and native wp_mysql_parser verified. Result: 266.32s. That is slower than current trunk evidence (100.28s), #235 (85.41s), and the #235 + #239 stack (56.86s).

PR #226, opened 2026-05-20
Use native plain text literal URL rewrites

adamziel:
## What changed

Uses the optional `wp_native_apis_rewrite_plain_text_literal_urls()` function when the php-toolkit native extension exposes it.

The existing PHP literal fast path remains the fallback. If the native function is missing, disabled, or returns `false`, reprint continues through the same parser-owned PHP path as before.

Depends on WordPress/php-toolkit#285 for the native PHP.wasm function.

## Why

The PHP literal fast path removed the large URL parser cost, but still handled 251k plain leaf rewrites in userland on the large fixture. This moves that narrow leaf rewrite into native code without changing the rigor boundary: structured data still has to be parsed by its owning parser, and the native function refuses ambiguous text.

## Benchmarks

Fixture: 262 MB SQL dump, 865 statements, 928,540 base64 values, 251,258 rewritten source-domain values. PHP.wasm 8.4.21, sqlite parser native extension loaded. Native run uses the locally built manifest from WordPress/php-toolkit#285.

### Rewrite-only profile

| path | total profile | plain literal rewrite bucket | change |
|---|---:|---:|---:|
| PHP literal fallback | 7.53 s | 1.86 s | baseline |
| Native literal function | 6.45 s | 0.83 s | 1.17x total, 2.25x bucket |

### Full db-apply

| path | run 1 | run 2 | mean |
|---|---:|---:|---:|
| PHP literal fallback | 31.60 s | - | 31.60 s |
| Native literal function | 28.80 s | 30.12 s | 29.46 s |

That is roughly 6.8% faster end-to-end on this fixture. The smaller full-run delta is expected: after the previous fast path, URL rewriting is no longer the dominant cost and the remaining SQLite/import floor is large.

## Validation

- `composer test -- UrlRewriting/StructuredDataUrlRewriterTest.php UrlRewriting/SqlStatementRewriterTest.php`
- `composer analyze`
- `composer build:phar`
- Full PHP.wasm db-apply with old php-toolkit manifest
- Full PHP.wasm db-apply with the locally built php-toolkit native literal manifest
- Rewrite-only PHP.wasm profiles for both manifests

### DIFF
--- a/packages/reprint-importer/src/lib/url-rewrite/class-structured-data-url-rewriter.php
+++ b/packages/reprint-importer/src/lib/url-rewrite/class-structured-data-url-rewriter.php
@@ -46,6 +46,12 @@ class StructuredDataUrlRewriter
      */
     private array $plain_text_literal_origin_mappings;
 
+    /** @var string Compact source-origin to target-prefix mappings for the native fast path. */
+    private string $native_plain_text_literal_origin_mapping;
+
+    /** @var bool Whether the native plain text literal URL rewrite function is available. */
+    private bool $native_plain_text_literal_url_rewriter_available;
+
     /**
      * Pre-parsed url_mapping grouped by the URL origin fields used by
      * is_child_url_of(): protocol and normalized hostname. Each entry is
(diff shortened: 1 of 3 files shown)


adamziel:
Closing under the current accepted full db-apply gate. This PR is stacked on #225, whose branch now measures 266.32s with the accepted playground-sqlite-db-apply gate when only the native MySQL parser manifest is enabled. The native literal extension path is not part of the current verified stack, and the surviving #235 + #239 stack is 56.86s.

PR #227, opened 2026-05-20
Restore structured URL rewrite ownership

adamziel:
## What it does

- Removes the literal base-URL replacement shortcut from `StructuredDataUrlRewriter`.
- Keeps values known to be block markup on the block-markup parser path.
- Updates comments and tests that described plain-text URL rewriting as `strtr()`.

## Rationale

The removed shortcut rewrote URL bytes before the structured URL processors owned the value. That was fast, but it violated the parser-ownership rule: block markup, block-comment JSON, HTML attributes, and plain text should each be rewritten by the processor responsible for that format.

This PR intentionally establishes the rigorous baseline for the follow-up speed experiments.

## Benchmark

|---|---:|---:|---:|
| `trunk` -> this PR | 139.06 s | 155.27 s | +16.21 s / +11.7% |

The regression is expected: this removes an unsafe shortcut and restores parser-owned rewriting. Follow-up PRs in the stack try to recover the performance without reintroducing literal URL replacement.

## Testing

- `vendor/bin/phpunit --configuration tests/phpunit.xml tests/UrlRewriting`

### DIFF
--- a/packages/reprint-importer/src/lib/url-rewrite/class-sql-statement-rewriter.php
+++ b/packages/reprint-importer/src/lib/url-rewrite/class-sql-statement-rewriter.php
@@ -12,7 +12,7 @@
  * WordPress core columns known to contain block markup (post_content, comment_content,
  * etc.) get the 'block_markup' hint so wp_rewrite_urls() handles HTML attributes,
  * block comment JSON, and CSS url(). All other columns default to auto-detect with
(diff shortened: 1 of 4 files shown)


adamziel:
Closing under the current db-apply rule. The PR body records a full db-apply regression versus trunk (139.06s -> 155.27s, +11.7%), so this does not survive the full-phase scrutiny.

PR #228, opened 2026-05-20
Explore exact structured URL rewrite cache

adamziel:
## What it does

- Adds a bounded exact-value cache around decoded structured URL rewrites.
- Keys cached results by content type and exact input bytes, with the hash used only as an array key.
- Checks the stored input again before returning a cached result, so hash collisions fall back to a miss.
- Caps cached entry size and entry count to avoid unbounded memory growth.

## Rationale

This explores whether repeated decoded leaves are a meaningful cost after restoring parser-owned rewriting. The cache is deliberately exact: it does not infer URL boundaries, trim delimiters, or reinterpret container syntax.

## Benchmark

|---|---:|---:|---:|
| previous PR -> this PR | 155.27 s | 158.91 s | +3.64 s / +2.3% |

This did not improve the large fixture. Keeping it as a separate draft makes the negative result explicit and keeps the cache decision isolated from the SQLite execution win in the next PR.

## Testing

- `vendor/bin/phpunit --configuration tests/phpunit.xml tests/UrlRewriting/StructuredDataUrlRewriterTest.php tests/UrlRewriting/SqlStatementRewriterTest.php`

### DIFF
--- a/packages/reprint-importer/src/lib/url-rewrite/class-structured-data-url-rewriter.php
+++ b/packages/reprint-importer/src/lib/url-rewrite/class-structured-data-url-rewriter.php
@@ -31,6 +31,8 @@ class StructuredDataUrlRewriter
     const BLOCK_MARKUP = 'block_markup';
     const PLAIN_TEXT = 'plain_text';
     private const REWRITE_RESULT_CACHE_MAX = 4096;
+    private const VALUE_REWRITE_CACHE_MAX = 2048;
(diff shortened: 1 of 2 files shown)


adamziel:
Closing under the current db-apply rule. The PR body records this exact-value cache as slower than its parent in full db-apply (155.27s -> 158.91s, +2.3%), so it is a rejected idea rather than a useful speedup PR.

PR #229, opened 2026-05-20
Inline base64 values for SQLite db apply

adamziel:
## What it does

- Rewrites complete `FROM_BASE64(...)` expressions to SQLite-adapter-compatible MySQL hex literals during SQLite `db-apply`.
- Reuses the existing scanner when URL rewriting is already walking the statement.
- Adds a standalone no-mapping path so SQLite imports avoid `FROM_BASE64` callbacks even when no URL mapping is configured.
- Keeps the registered SQLite `FROM_BASE64` function as a fallback for statements that are not inlined and for post-apply maintenance queries.

## Rationale

The SQLite adapter was spending substantial time invoking a PHP user-defined function for every encoded value. The importer has already scanned the SQL statement, and the dump's encoded payloads are base64, so it can decode them once and hand the SQLite adapter a MySQL hex literal (`0x...`) instead of a function call.

Empty decoded strings use `''` because bare `0x` is not a valid MySQL token. Non-empty values use hex literals so arbitrary bytes do not enter the SQL text as quoted strings.

## Benchmark

|---|---:|---:|---:|
| previous PR -> this PR | 158.91 s | 125.12 s | -33.79 s / -21.3% |
| `trunk` -> full stack | 139.06 s | 125.12 s | -13.94 s / -10.0% |

This is the only branch in the stack that produced a meaningful speedup on the large fixture.

## Testing

- `vendor/bin/phpunit --configuration tests/phpunit.xml tests/UrlRewriting tests/Import/NewSiteUrlSqliteTest.php`

### DIFF
--- a/packages/reprint-importer/src/import.php
+++ b/packages/reprint-importer/src/import.php
@@ -5118,6 +5118,7 @@ class ImportClient
         }
 
         [$pdo, $connection_label] = $this->create_target_db_apply_connection($options);
+        $inline_base64_values_for_sqlite = strtolower((string) ($options["target_engine"] ?? "mysql")) === "sqlite";
 
         $this->audit_log(
             "CONNECTED | {$connection_label}",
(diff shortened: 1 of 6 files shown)


adamziel:
Closing under the current full db-apply review. This PR is stacked on #228, which is now closed because it slowed full db-apply, and its recorded full-stack result (125.12s) is superseded by the current #235 + #239 stack (56.86s on the accepted playground-sqlite-db-apply gate, native parser verified).

PR #233, opened 2026-05-21
[codex] Fast path text-only block markup rewrites

adamziel:
## What it does

Adds a structured fast path for `post_content` values known by the SQL layer to be block markup when every source-domain occurrence is in text, not inside tag/comment markup.

- Keeps structured URL handling through `URLInTextProcessor`, `WPURL::parse()`, `is_child_url_of()`, and `WPURL::replace_base_url()`.
- Keeps full `BlockMarkupUrlProcessor` handling for HTML attributes, block-comment JSON, CSS `url()`, IDNs, case variants, and structured envelopes.
- Adds bounded whole-value memoization for repeated text-node-only block-markup values, guarded by exact input-byte recheck before cache reuse.
- Adds regressions for text-node rewriting and for leaving source-domain text inside an unknown HTML attribute untouched.

No literal URL `str_replace`/`strtr` path is introduced.

## Benchmark

Focused PHP microbenchmark against PR #231 base for repeated text-node-only paragraph block markup:

| Case | Time | Delta |
|---|---:|---:|
| PR #231 base | 297.96 ms | baseline |
| this branch | 10.633 ms | 28.0x faster |

Worker run also measured:

| Case | Before | After | Delta |
|---|---:|---:|---:|
| Cold unique plain post content, 1200 values | 347.87 ms | 270.27 ms | 1.29x |
| Cold repeated paragraph block markup, 1200 values | 297.96 ms | 7.38 ms | 40.37x |
| Cold unique structured attributes, 500 values | 660.33 ms | 685.05 ms | 0.96x |

The large win is intentionally scoped to repeated text-node-only block markup. Structured HTML attributes stay on the full parser path and are roughly neutral/slightly slower in the microbench due to the extra prefilter.

## Testing

```bash
./vendor/bin/phpunit --configuration tests/phpunit.xml tests/UrlRewriting
./vendor/bin/phpstan analyze --memory-limit=1G packages/reprint-importer/src/lib/url-rewrite/class-structured-data-url-rewriter.php
```

Result: 214 tests, 2282 assertions, 7 skipped; PHPStan passed.

### DIFF
--- a/packages/reprint-importer/src/lib/url-rewrite/class-structured-data-url-rewriter.php
+++ b/packages/reprint-importer/src/lib/url-rewrite/class-structured-data-url-rewriter.php
@@ -236,6 +236,7 @@ class StructuredDataUrlRewriter
 
         if (
             !$looks_like_structured_envelope &&
+            !$this->contains_raw_text_element($value) &&
             $this->source_domain_occurrences_are_outside_markup($value)
         ) {
             if (strpos($value, '<') === false) {
@@ -256,6 +257,14 @@ class StructuredDataUrlRewriter
         return $looks_like_structured_envelope
             ? $this->rewrite($value, self::BLOCK_MARKUP)
             : $this->rewrite_urls($value, self::BLOCK_MARKUP);
+    }
+
+    private function contains_raw_text_element(string $value): bool
+    {
+        return stripos($value, '<script') !== false
+            || stripos($value, '<style') !== false
+            || stripos($value, '<textarea') !== false
+            || stripos($value, '<title') !== false;
     }
 
     private function looks_like_structured_envelope(string $value): bool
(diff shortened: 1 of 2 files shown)


adamziel:
Post-scrutiny verification after adding the raw-text-element guard:

- Adversarial test added: block text fast path must not rewrite URLs inside `<script>` / `<textarea>`, while still rewriting normal text-node URLs.
- Focused benchmark still clears the bar: PR #231 base median `3214.275ms`, current branch median `76.105ms`, about `42.2x`.
- Focused PHPUnit passed: `StructuredDataUrlRewriterTest.php`.

adamziel:
Closing after full database-apply evaluation. Microbenchmarks are no longer considered valid evidence for this work. Current full  benchmark with native parser verified: current trunk ; this branch rebased onto current trunk . This does not speed up the actual database apply phase and makes it slower.

adamziel:
Correction to the close note: full database-apply evidence was trunk 100.28s versus this branch 121.18s for the full playground-sqlite-db-apply stage with native parser verified. The branch is closed because it slows down the actual database apply phase.

PR #234, opened 2026-05-21
[codex] Scan sparse URL rewrite candidates

adamziel:
## Summary

Keeps the sparse URL candidate scanner only because it has a measured full database-apply impact. The scanner narrows URL rewriting work to FROM_BASE64 payloads whose encoded bytes can contain an http/https prefix, while preserving fallback behavior for unsupported shapes.

This is evaluated only on the full database apply phase. Microbenchmarks are not used as evidence.

## Full db-apply benchmark

Both runs use `BENCH_STAGES=playground-sqlite-db-apply`, PHP.wasm 8.3, and the native `wp_mysql_parser` extension manifest.

| Branch | playground-sqlite-db-apply |
| --- | ---: |
| current trunk | 100.28 s |
| this branch, rebased on trunk | 89.78 s |

Native parser verification was enabled in the benchmark output.

## Verification

- `./vendor/bin/phpunit --configuration tests/phpunit.xml tests/UrlRewriting/SqlStatementRewriterPrefilterTest.php`
- Result: 28 tests, 1927 assertions.

### DIFF
--- a/packages/reprint-importer/src/lib/url-rewrite/class-base64-value-scanner.php
+++ b/packages/reprint-importer/src/lib/url-rewrite/class-base64-value-scanner.php
@@ -37,8 +37,9 @@ class Base64ValueScanner
      *   'encoded_value'=> string The base64 payload
      *   'value'        => ?string The base64-decoded value, cached on demand
      *   'new_value'    => ?string Non-null when set_value() has been called
+     *   'column_name'  => ?string Optional column name from tokenization-free scanners
      *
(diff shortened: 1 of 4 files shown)


adamziel:
Closing after full database-apply evaluation against the current best candidate. Microbenchmarks are not valid evidence for this work. Full playground-sqlite-db-apply timings with native parser verified: trunk 100.28s; this PR alone 89.78s; clean #235 alone 74.83s; #234 + clean #235 combined 107.85s. Since it makes the best full-db-apply candidate worse than trunk when combined, it should not remain open as an optimization PR.

PR #235, opened 2026-05-21
[codex] Skip block text scans for markup-only URL hits

adamziel:
## What it does

Skips block text-node URL scanning when raw source-domain hits are confined to closed HTML tag/comment markup, while preserving structured block/HTML rewriting.

## Rationale

The original goal is full `playground-sqlite-db-apply` speed, not method-level timing. Avoiding unnecessary text scans helps the actual SQLite apply phase, but the fast path must not skip text URLs that only become visible after entity decoding or after literal angle-bracket text.

Full-stage evidence for the current branch:

- `trunk`: `100.28s`
- this branch after the correctness hardening: `85.41s`

Both timings are full `BENCH_STAGES=playground-sqlite-db-apply` runs with PHP 8.3, `PLAYGROUND_PHP_USE_WASM_RUNNER=1`, and native `wp_mysql_parser` verified.

## Implementation

- Treats only closed, likely tag/comment spans as markup for the skip decision.
- Checks raw and HTML entity-decoded text spans before skipping text-node scanning.
- Adds adversarial coverage for literal `< ... >` text and entity-decoded text URLs.

## Testing instructions

```bash
vendor/bin/phpunit --configuration tests/phpunit.xml tests/UrlRewriting/StructuredDataUrlRewriterTest.php
BENCH_STAGES=playground-sqlite-db-apply \
BENCH_LABEL=pr235-supervised-correctness-fix \
WP_MYSQL_PARSER_EXTENSION_MANIFEST=https://wordpress.github.io/sqlite-database-integration/wp_mysql_parser-wasm-extension/latest/manifest.json \
PLAYGROUND_PHP_USE_WASM_RUNNER=1 \
PLAYGROUND_PHP_VERSION=8.3 \
node tests/e2e/benchmark/bench-pull.mjs
```

Observed test result: `53 tests`, `84 assertions`, `7 skipped`.
Observed full db-apply result: `85.41s`, native parser verified.

PR #237, opened 2026-05-21
[codex] Cache SQLite prepared insert shapes

adamziel:
## Summary

Stacked on #235. This caches the parsed producer INSERT shape in `SQLitePreparedInsertBuilder`, so repeated table/column/value layouts can skip the full `FastInsertScanner` pass while still decoding each current FROM_BASE64 payload and running the structured rewrite callback with the current table and column.

The cached path validates literal slots, numeric slots, wrapper syntax, statement length, and strict base64 decoding before using the shape. If the current SQL does not match, it falls back to the normal scanner path.

## Adversarial scrutiny

- Added coverage proving dynamic numeric slots and per-value columns remain current across cache hits.
- Added coverage for malformed numeric slots like `1e`, which must not be accepted by the cached fast path.
- Tightened the shared scanner numeric parser as well, so the fallback path and cached path fail consistently.
- Removed duplicate regex base64 validation from the cached path and now fail closed on strict `base64_decode(..., true)` instead.

## Benchmark

Builder-only benchmark using 30,000 precomputed producer-shaped INSERTs with dynamic numeric IDs and two FROM_BASE64 payload columns:

| branch | median | speed-up |
| --- | ---: | ---: |
| `origin/codex/pr231-skip-block-text-scan` | 287.941 ms | 1.00x |
| `codex/pr235-sqlite-shape-cache` | 138.708 ms | 2.08x |

This intentionally excludes SQL generation time and measures the hot builder path directly.

## Checks

- `./vendor/bin/phpunit --configuration tests/phpunit.xml tests/UrlRewriting/SQLitePreparedInsertBuilderTest.php`
- `./vendor/bin/phpunit --configuration tests/phpunit.xml tests/UrlRewriting`
- `./vendor/bin/phpstan analyze --memory-limit=1G packages/reprint-importer/src/lib/url-rewrite/class-sqlite-prepared-insert-builder.php packages/reprint-importer/src/lib/url-rewrite/class-fast-insert-scanner.php`

### DIFF
--- a/packages/reprint-importer/src/lib/url-rewrite/class-fast-insert-scanner.php
+++ b/packages/reprint-importer/src/lib/url-rewrite/class-fast-insert-scanner.php
@@ -37,7 +37,8 @@ class FastInsertScanner
      * @return array{
      *   table: string,
      *   column_map: list<array{int, int, string}>,
-     *   base64_entries: list<array{expr_start: int, expr_length: int, quote_start: int, quote_length: int, encoded_value: string, value: ?string, new_value: ?string}>
+     *   base64_entries: list<array{expr_start: int, expr_length: int, quote_start: int, quote_length: int, encoded_value: string, value: ?string, new_value: ?string}>,
+     *   value_entries: list<array{start: int, length: int, kind: string, column: string, quote_start?: int, quote_length?: int, encoded_value?: string}>
      * }|null
      *   Null when the SQL doesn't match the recognised shape.
      */
@@ -93,6 +94,7 @@ class FastInsertScanner
 
         $column_map = [];
         $base64_entries = [];
+        $value_entries = [];
 
         $cursor = $values_end;
         $sql_len = strlen($sql);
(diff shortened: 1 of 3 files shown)


adamziel:
Post-restack verification after rebasing onto the updated #235 branch:

- Full URL rewriting suite passed: `219 tests`, `2300 assertions`, `7 skipped`.
- PHPStan passed on the touched SQLite prepared insert builder and fast insert scanner files.
- Current builder benchmark: #235 base median `346.612ms`, current branch median `168.486ms`, about `2.06x`.
- Adversarial malformed numeric-slot coverage remains in place (`1e` must not be accepted by the cached shape path).

adamziel:
Closing after full database-apply evaluation. Microbenchmarks are no longer considered valid evidence for this work. Current full playground-sqlite-db-apply timings with native parser verified: trunk 100.28s; #235 stack 77.97s; this branch on top of that stack 90.33s. Although it is still faster than trunk in this run, it makes the best full-db-apply candidate materially slower, so it should not remain open as an optimization PR.

PR #239, opened 2026-05-21
[codex] Cache SQLite prepared insert shapes

adamziel:
## What it does

Caches the parsed value layout for producer-shaped SQLite prepared `INSERT` statements. Repeated batches with the same table/column/value shape can skip the full `FastInsertScanner` pass and rebuild prepared SQL from cached chunks.

## Rationale

The accepted gate is the full `playground-sqlite-db-apply` phase, not method-level timing. The prior stack included #235, but a fresh full-stage audit showed #235 does not improve db-apply. This PR has been rebased directly onto `trunk` and is the surviving db-apply speedup.

Full-stage evidence:

- fresh `trunk` audit: `73.86s`
- this PR rebased on `trunk`: `65.03s`

Both timings are full `BENCH_STAGES=playground-sqlite-db-apply` runs on `large-directory` with 320,007 posts / 720,015 postmeta, PHP 8.3 wasm runner, `PLAYGROUND_PHP_USE_WASM_RUNNER=1`, and native `wp_mysql_parser` verified. The PR run printed the completed full-stage table, then exited nonzero only because `.context/` did not exist for the JSON output path.

## Implementation

- Adds a bounded static shape cache to `SQLitePreparedInsertBuilder`.
- Stores literal chunks, static value kinds, base64 positions, table name, and column names from known-good producer-shaped inserts.
- Revalidates cached matches against the whole SQL string before using them.
- Adds a decoded-payload rewrite prefilter so non-URL payloads avoid the rewrite callback.
- Hardens invalid base64 handling so cached and uncached paths both fall back instead of binding empty strings.

## Testing instructions

```bash
vendor/bin/phpunit --configuration tests/phpunit.xml tests/UrlRewriting
mkdir -p .context
BENCH_STAGES=playground-sqlite-db-apply \
BENCH_LABEL=pr239-shape-on-trunk \
BENCH_JSON_OUT=.context/pr239-shape-on-trunk.json \
BENCH_MD_OUT=.context/pr239-shape-on-trunk.md \
WP_MYSQL_PARSER_EXTENSION_MANIFEST=https://wordpress.github.io/sqlite-database-integration/wp_mysql_parser-wasm-extension/latest/manifest.json \
PLAYGROUND_PHP_USE_WASM_RUNNER=1 \
PLAYGROUND_PHP_VERSION=8.3 \
node tests/e2e/benchmark/bench-pull.mjs
```

Observed URL-rewrite suite on the rebased branch: `225 tests`, `2303 assertions`, `7 skipped`.
Observed full db-apply result on the rebased branch: `65.03s`, native parser verified.

### DIFF
--- a/packages/reprint-importer/src/lib/url-rewrite/class-fast-insert-scanner.php
+++ b/packages/reprint-importer/src/lib/url-rewrite/class-fast-insert-scanner.php
@@ -37,7 +37,17 @@ class FastInsertScanner
      * @return array{
      *   table: string,
      *   column_map: list<array{int, int, string}>,
-     *   base64_entries: list<array{expr_start: int, expr_length: int, quote_start: int, quote_length: int, encoded_value: string, value: ?string, new_value: ?string}>
+     *   base64_entries: list<array{
+     *     expr_start: int,
+     *     expr_length: int,
+     *     quote_start: int,
+     *     quote_length: int,
+     *     encoded_value: string,
+     *     value: ?string,
+     *     new_value: ?string,
+     *     column: string
+     *   }>,
+     *   value_entries: list<array{start: int, length: int, kind: string, column: string, quote_start?: int, quote_length?: int, encoded_value?: string}>
      * }|null
      *   Null when the SQL doesn't match the recognised shape.
      */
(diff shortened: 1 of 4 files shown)


adamziel:
Closing the bundled PR so this pass can split it into separate single-optimization PRs. The combined branch mixed prepared INSERT shape caching, decoded-payload prefiltering, and base64 hardening, which made it hard to scrutinize each effect against the full db-apply gate. New worker lanes are splitting these pieces and will only keep PRs that survive full playground-sqlite-db-apply evidence.

PR #241, opened 2026-05-21
Batch SQLite db-apply work between state checkpoints

adamziel:
## What it does

Batches SQLite `db-apply` work inside explicit SQLite transactions aligned with the existing 100-statement state checkpoints.

## Rationale

`SET AUTOCOMMIT=0` in the dump only updates MySQL-on-SQLite session state; it does not keep the SQLite connection in one write transaction. The existing state-save boundary is the durability boundary this code can safely resume from.

## Implementation

- Starts a SQLite transaction before applying a checkpoint window.
- Commits before each periodic `save_state()` checkpoint, then starts the next window.
- Commits before graceful partial state saves and before final plugin deactivation/state completion.
- Rolls back the current uncheckpointed window on apply errors so resume replays from the last committed byte boundary.

## Full db-apply evidence

Performance Report run: https://github.com/adamziel/reprint/actions/runs/26226894929/job/77176319597

`large-directory` benchmark, with 10,000 posts, 25,000 postmeta rows, 2,000+ files, PHP 8.5.6 runner, and Playground PHP.wasm 8.3:

- PR `playground-sqlite-db-apply`: `3.67 s`, `1` resume attempt, status `✓`.
- Trunk `playground-sqlite-db-apply`: `3.82 s`, `1` resume attempt, status `✓`.
- Details: `condition=db-apply to SQLite in PHP.wasm`, `wp_mysql_parser=enabled`, `mode=parser`, `native_lexer=verified`, `native_parser=verified`, `native_ast=WP_MySQL_Native_Parser_Node`, `sqlite_driver_parser=verified`.

## Transaction/resume risk

This changes SQLite `db-apply` durability from statement-by-statement autocommit to checkpoint-sized transactions. A failure inside a checkpoint window rolls back that window and resume replays from the last saved SQL byte offset. A crash after SQLite commit but before the state file is saved can still leave committed database work ahead of the saved resume offset, so commit and state save are not atomic across storage layers.

### DIFF
--- a/packages/reprint-importer/src/import.php
+++ b/packages/reprint-importer/src/import.php
@@ -5135,7 +5135,7 @@ class ImportClient
             if (!$use_sqlite_apply_transactions || $sqlite_apply_transaction_active) {
                 return;
             }
-            if (method_exists($pdo, 'inTransaction') && $pdo->inTransaction()) {
+            if ($pdo->inTransaction()) {
                 return;
             }
             $pdo->beginTransaction();
@@ -5149,7 +5149,7 @@ class ImportClient
             if (!$use_sqlite_apply_transactions || !$sqlite_apply_transaction_active) {
                 return;
             }
-            if (method_exists($pdo, 'inTransaction') && $pdo->inTransaction()) {
+            if ($pdo->inTransaction()) {
                 $pdo->commit();
             }
             $sqlite_apply_transaction_active = false;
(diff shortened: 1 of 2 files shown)


adamziel:
Closing because this does not meet the current bar for this pass. The accepted evidence is the full database apply phase, and the measured improvement is too small to matter for the import target. We are pivoting the worker lanes to structured-data-safe ideas that can plausibly cut full db-apply time by 50% or more.

PR #244, opened 2026-05-21
Add structured SQLite insert row streaming

adamziel:
## What it does

Adds a structured SQLite `INSERT` path for producer-shaped row streams. Supported rows are parsed into a stable, fully parameterized SQLite statement, then executed through a small prepared-statement cache during `db-apply`.

Unsupported shapes still return `null` and use the existing SQL-text fallback.

## Tokenizer cleanup

The structured SQLite insert builder now uses `WP_MySQL_Lexer` tokens for the supported SQL shape instead of regex/manual SQL-shape parsing. The token walk handles table names, column lists, row tuples, numeric literals, empty strings, `NULL`, `FROM_BASE64(...)`, and `CONVERT(FROM_BASE64(...) USING utf8mb4)` as structured tokens.

Correctness remains the priority: unsupported or ambiguous shapes fall back safely rather than being guessed from raw string positions. The structured path still calls the rewrite callback for accepted base64 values with table and column context.

## Adversarial coverage

Added/kept focused tests for:

- comments containing fake `FROM_BASE64(...)` before and inside the statement
- escaped backtick identifiers in table and column names
- string payloads containing `),(` and fake `FROM_BASE64(...)` text
- `CONVERT(FROM_BASE64(...) USING utf8mb4)` values
- malformed/non-base64 payloads
- nested expressions such as `(1 + 2)`, `ABS(-1)`, `CONCAT(FROM_BASE64(...), ...)`
- unsupported `CONVERT('plain' USING utf8mb4)` fallback
- unsupported shapes continuing to return `null`

## Benchmark

Preliminary full db-apply benchmark from the tokenizer-cleanup worktree:

```bash
BENCH_STAGES=playground-sqlite-db-apply \
WP_MYSQL_PARSER_EXTENSION_MANIFEST=https://wordpress.github.io/sqlite-database-integration/wp_mysql_parser-wasm-extension/latest/manifest.json \
PLAYGROUND_PHP_USE_WASM_RUNNER=1 \
PLAYGROUND_PHP_VERSION=8.3 \
node tests/e2e/benchmark/bench-pull.mjs
```

Result:

- `playground-sqlite-db-apply`: **96.96s**, attempts: 1, status: ok
- Runtime: `php.wasm 8.3`
- Native parser evidence: `wp_mysql_parser=enabled`, native lexer/parser verified, SQLite driver native parser verified

Caveat: this timing is preliminary only. The run overlapped active #243/#245/#246 benchmark processes, so it should not be used as a clean performance comparison. Re-run in a quiet window before making performance claims.

## Testing

Focused verification after tokenizer cleanup:

```bash
php -l packages/reprint-importer/src/lib/url-rewrite/class-sqlite-prepared-insert-builder.php
php -l tests/UrlRewriting/SQLitePreparedInsertBuilderTest.php
./vendor/bin/phpunit tests/UrlRewriting/SQLitePreparedInsertBuilderTest.php
./vendor/bin/phpunit tests/UrlRewriting
```

Results:

- `SQLitePreparedInsertBuilderTest`: 16 tests, 81 assertions
- `tests/UrlRewriting`: 231 tests, 2359 assertions, 7 skipped

### DIFF
--- a/packages/reprint-importer/src/import.php
+++ b/packages/reprint-importer/src/import.php
@@ -5119,6 +5119,7 @@ class ImportClient
 
         [$pdo, $connection_label] = $this->create_target_db_apply_connection($options);
         $sqlite_prepared_pdo = null;
+        $sqlite_prepared_statement_cache = [];
         if (
             strtolower((string) ($options["target_engine"] ?? "mysql")) === "sqlite"
             && method_exists($pdo, 'get_connection')
@@ -5245,6 +5246,7 @@ class ImportClient
                             $query,
                             $stmt_rewriter,
                             $sqlite_prepared_pdo,
+                            $sqlite_prepared_statement_cache,
                             $executed_query,
                         );
                     } catch (PDOException $e) {
@@ -5323,6 +5325,7 @@ class ImportClient
                         $query,
                         $stmt_rewriter,
                         $sqlite_prepared_pdo,
+                        $sqlite_prepared_statement_cache,
                         $executed_query,
                     );
                 } catch (PDOException $e) {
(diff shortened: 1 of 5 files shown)


adamziel:
Closing after replacing the regex/manual SQL-shape parsing with lexer-token based structured insert parsing and re-benchmarking on current trunk. The safer tokenizer version passed URL rewriting tests and parser/native SQLite verification, and PR #244 is no longer conflicted, but clean db-apply is 81.74s. That is slower than the surviving PR #243 at 61.38s, so this row-stream lane no longer survives the performance bar.

PR #255, opened 2026-05-22
Optimize resumed db-pull SQL file writes

adamziel:
## What it does

Speeds up PHP.wasm `db-pull` file output by changing resumed `db.sql` writes from append mode to a single seek followed by sequential writes. It also gives the SQL file stream a 1 MiB write buffer and keeps `sql_statements_counted` in the normalized state so partial/resumed pulls preserve statement-count progress.

## Rationale

The PHP.wasm runner writes `db.sql` through a mounted filesystem. Append-mode writes are more expensive there because every write takes the append path. The importer already knows the checkpointed SQL byte count, so it can seek to the expected end once on resume and write normally.

Benchmark under `flock /tmp/reprint-bench.lock -c ...` with `BENCH_STAGES=playground-sqlite-db-pull`:

| Version | Wall time | Delta |
|---|---:|---:|
| `origin/trunk` | 441.66 s | baseline |
| this branch | 405.16 s | -36.50 s (-8.3%) |

## Implementation

On file-mode `download_sql()`:

- Reuses the initial `filesize()` result during crash recovery instead of restatting.
- Opens resumed `db.sql` with `c` and calls `fseek($sql_bytes_written)` once instead of opening with `a`.
- Refuses resume when `db.sql` is missing or shorter than the saved SQL byte checkpoint, avoiding silent holes after local file loss or truncation.
- Uses `stream_set_write_buffer()` with a 1 MiB buffer.
- Leaves SQL chunk bytes, cursors, domain discovery, and error handling paths intact.
- Adds `sql_statements_counted` to the default state schema so existing save/resume code is not discarded by normalization.

## Testing instructions

```bash
php -l tests/Import/SqlFileWritePathTest.php
./vendor/bin/phpunit --configuration tests/phpunit.xml tests/Import/SqlFileWritePathTest.php
./vendor/bin/phpunit --configuration tests/phpunit.xml tests/Import/CurlTimeoutRecoveryTest.php
git diff --check
./vendor/bin/phpstan analyse --configuration=phpstan.neon --memory-limit=1G packages/reprint-importer/src/import.php
flock /tmp/reprint-bench.lock -c 'BENCH_STAGES=playground-sqlite-db-pull node tests/e2e/benchmark/bench-pull.mjs'
```

### DIFF
--- a/packages/reprint-importer/src/import.php
+++ b/packages/reprint-importer/src/import.php
@@ -7180,7 +7180,22 @@ class ImportClient
                         fclose($handle);
                     }
                     $actual_size = $tracked_bytes;
+                } elseif ($actual_size < $tracked_bytes) {
+                    throw new RuntimeException(
+                        sprintf(
+                            "Cannot resume SQL download: db.sql is smaller than the saved checkpoint (%d < %d bytes). Use --abort to restart db-pull.",
+                            $actual_size,
+                            $tracked_bytes,
+                        )
+                    );
                 }
+            } elseif ($cursor && $tracked_bytes !== null && $tracked_bytes > 0) {
+                throw new RuntimeException(
+                    sprintf(
+                        "Cannot resume SQL download: db.sql is missing but the saved checkpoint expects %d bytes. Use --abort to restart db-pull.",
+                        $tracked_bytes,
+                    )
+                );
             }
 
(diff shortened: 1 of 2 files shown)


adamziel:
sounds like risk of not writing what we wanted to write in case of crash

PR #258, opened 2026-06-11
Prevent errors during compressibility checks

WPprodigy:
The per-batch gzip compressibility check tries to call `fread()` on every entry, including directories which causes a fatal error. The `fopen()` succeeds but then `fread()` fatals, crashing the whole export mid files-sync with:

> PHP Error: fread(): Read of 8192 bytes failed with errno=21 Is a directory

Fix:
- Guard with `is_file()`,  skipping any non-regular path (directory, symlink, missing) instead of trying to read it.
- Performance improvement: Short-circuit the checking once the batch is already known compressible, skipping the remaining classification/byte-sniffing.

## Testing

Try a migration when you have a theme folder named something like `storefront-1.2.3` (a period somewhere in the name).

### DIFF
--- a/packages/reprint-exporter/src/export.php
+++ b/packages/reprint-exporter/src/export.php
@@ -3448,6 +3448,10 @@ function file_fetch_paths_should_gzip(array $paths): bool
             // shouldn't compress around. Treat as a hard reject.
             return false;
         }
+        // Once true, we can skip checking the subsequent files.
+        if ($any_compressible) {
+            continue;
+        }
         $ext = path_extension_compressibility($path);
         if ($ext === 'yes') {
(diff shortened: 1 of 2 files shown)


chrisbliss18 (changes requested):
The fix looks good. I found that it fixes both passing a dotted directory name like `storefront-2.4.5` and a symlink with such a name to a directory as an argument to `runFileFetch()`.

Your test adds a check for the directory name issue but not the symlink one. Could you expand the test or add a new one to test the symlink issue?

chrisbliss18 (approved):
Okay. I was thinking about potential regressions in case the logic changed later. That said, I'm good with the changes.

Approved.

PR #260, opened 2026-06-15
Refresh on re-run for completed pull sub-commands

epeicher:
## Problem

Re-pulling a site never propagated **edits or deletions to existing database rows** — the local database stayed frozen at the content captured by the first completed pull. The root cause is an asymmetry in the engine: the composite `pull` orchestrator resets state for a delta re-pull (`Pull::prepare_repull()`), but callers that drive the sub-commands directly (notably Studio's `pull-reprint`) had no refresh path:

| Sub-command | Re-run on completed state (before) |
| --- | --- |
| `files-pull` | Silent no-op ("already complete… use `--abort`") |
| `db-pull` | Throws "already completed … Use --abort flag to start over." |
| `db-apply` | Throws "already completed. Use --abort flag to re-run." |

`--abort` is destructive (re-downloads everything), so there was no clean incremental re-sync.

## Change

Re-invoking a sub-command on a **completed** state now performs a refresh, matching the composite `pull`'s re-pull semantics:

- **`files-pull` → delta sync.** Keeps the local index, re-indexes the remote, diffs, fetches only changes. Stale remote index / download lists are cleared first (`download_remote_index()` appends when the file exists).
- **`db-pull` → full refresh.** New `ImportClient::reset_db_pull_state()` (now also used by `Pull::prepare_repull()`) resets DB state and deletes `db.sql`, `.import-domains.json`, and the `.sql-buffer` crash-recovery file, then re-downloads the entire dump.
- **`db-apply` → re-apply from statement 0.** The dump is idempotent (`DROP TABLE IF EXISTS` + full INSERTs), so edits, inserts, and deletes all land.

In-progress/partial states still resume from their cursors; `--abort` still clears state without running.

## Bugs fixed along the way

1. **`--filter=skipped-earlier` never marked completion** — the branch returned before the `status = "complete"` marking, leaving every successful essential+skipped pull stuck `in_progress` forever. Fixed, plus a legacy-state repair in `run()` for sessions written by older versions.
2. **Deleting the skipped download list broke running runtimes** — `apply-runtime` mounts `state/.import-download-list-skipped.jsonl` into the generated runtime (remote-uploads proxy), and PHP-WASM re-resolves mounts on every runtime rotation. Deleting the file crashed server starts and WP-CLI with ENOENT. It is now truncated to empty instead (all has-skipped checks test `size > 0`, so semantics are unchanged).
3. **Phantom "X is currently editing" badges** — the dump carries the remote's ephemeral `_edit_lock` postmeta, which renders as an editor lock on the imported copy for 150s after the captured timestamp. `db-apply` now strips `_edit_lock` rows after applying (WordPress regenerates them locally; core's WXR exporter skips them too).
4. **Stale `.sql-buffer` replay** — a leftover crash-recovery buffer could be replayed into the target on a refresh (latent `prepare_repull()` gap, fixed via the shared reset helper).

## Known limitation

A table dropped on the remote between pulls lingers in the local target (the new dump no longer mentions it). WP core tables always exist, so this is rare; recreating the SQLite file on refresh is a possible hardening follow-up.

## Testing

- New `tests/Import/DbResyncStateTest.php`: round-trip via the sub-command path proving an **edited**, **inserted**, and **deleted** row all propagate on re-run; db-pull refresh resets state and deletes stale artifacts; the stuck skipped-earlier state is repaired; `_edit_lock` is stripped while other postmeta survives.
- `tests/Import/FilesSyncStateTest.php` updated: completed `files-pull` re-runs as a delta and clears the stale remote index.
- Full `tests/Import/` + `tests/UrlRewriting/` suites green (493 tests).
- Verified live end-to-end against a WP.com Atomic site via Studio: remote post edits, inserts, and deletes propagate on re-pull; site start and WP-CLI work after sync cycles.

---

_Supersedes #257 — moved from a fork to an in-repo branch so CI runs with full token permissions (the Performance Report perf-comment step can post on this PR)._

### DIFF
--- a/packages/reprint-importer/src/import.php
+++ b/packages/reprint-importer/src/import.php
@@ -1567,6 +1567,27 @@ class ImportClient
 
         $this->state = $this->load_state();
 
+        // Legacy state repair: older versions never marked a
+        // --filter=skipped-earlier run complete, leaving the state stuck at
+        // "in_progress" forever after a fully successful fetch (stage reset
+        // to null and the skipped download list deleted on completion).
+        // Detect that terminal shape and mark it complete so re-runs take
+        // the delta path instead of tripping the mid-flight filter guard.
+        if (
+            ($this->state["command"] ?? null) === "files-pull" &&
+            ($this->state["status"] ?? null) === "in_progress" &&
+            ($this->state["stage"] ?? null) === null &&
+            ($this->state["filter"] ?? null) === "skipped-earlier" &&
+            !file_exists($this->skipped_download_list_file)
+        ) {
+            $this->audit_log(
+                "STATE REPAIR | skipped-earlier run finished but was never marked complete — marking complete",
+                true,
+            );
+            $this->state["status"] = "complete";
+            $this->save_state($this->state);
+        }
+
         // Persist follow_symlinks in state so it survives across invocations.
         // If explicitly set on CLI, store it.  Otherwise, restore from persisted state.
         if (isset($options["follow_symlinks"])) {
@@ -1888,10 +1909,7 @@ class ImportClient
                     @unlink($this->download_list_file);
                     $this->audit_log("FILE DELETE | {$this->download_list_file}");
                 }
-                if (file_exists($this->skipped_download_list_file)) {
-                    @unlink($this->skipped_download_list_file);
-                    $this->audit_log("FILE DELETE | {$this->skipped_download_list_file}");
-                }
+                $this->clear_skipped_download_list("abort files-pull");
                 if (file_exists($this->volatile_files_file)) {
                     @unlink($this->volatile_files_file);
                     $this->audit_log("FILE DELETE | {$this->volatile_files_file}");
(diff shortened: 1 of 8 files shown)


packages/reprint-importer/src/import.php
```diff
@@ -1567,6 +1567,27 @@ public function run(array $options = []): void
 
         $this->state = $this->load_state();
 
+        // Legacy state repair: older versions never marked a
```
adamziel, line 1570: No need to document what older versions did inline. It only matters for reviewing this PR but will not matter once we merge it. Let's just document what this does now.

packages/reprint-importer/src/import.php
```diff
@@ -1567,6 +1567,27 @@ public function run(array $options = []): void
...
+        // the delta path instead of tripping the mid-flight filter guard.
+        if (
+            ($this->state["command"] ?? null) === "files-pull" &&
+            ($this->state["status"] ?? null) === "in_progress" &&
+            ($this->state["stage"] ?? null) === null &&
+            ($this->state["filter"] ?? null) === "skipped-earlier" &&
+            !file_exists($this->skipped_download_list_file)
+        ) {
+            $this->audit_log(
+                "STATE REPAIR | skipped-earlier run finished but was never marked complete — marking complete",
```
adamziel, line 1584: Shouldn't we mark it as complete right before deleting the `skipped_download_list_file`? As this stands now, a hypothetical premature deletion of the `skipped_download_list_file` would keep reprint running where I'd expect it to stop and communicate the error.

packages/reprint-importer/src/import.php
```diff
@@ -10188,6 +10270,68 @@ function ($ch, $dl_total, $dl_now, $ul_total, $ul_now) {
      * Reset state to defaults while preserving cross-command data like
      * preflight results, version, and follow_symlinks.
      */
+    /**
```
adamziel: Note the "Reset state" comment got misplaced now. There's also a misplaced comment before it already, let's find its home or delete it.

packages/reprint-importer/src/lib/pull/class-pull.php
```diff
@@ -397,39 +397,37 @@ private function normalize_url(): void
...
-            $state_dir . "/.import-domains.json",
             $state_dir . "/.import-remote-index.jsonl",
             $state_dir . "/.import-download-list.jsonl",
-            $state_dir . "/.import-download-list-skipped.jsonl",
         ] as $path) {
             if (file_exists($path)) {
                 @unlink($path);
             }
         }
+        // Truncated, not deleted: the generated runtime may hold a mount
```
adamziel, line 428: What does that mean? Is it for running in Studio? If so, let's clearly explain that scenario in the doc string. What's the setup? What's the problem? How is this solving the problem?

packages/reprint-importer/src/import.php
```diff
@@ -2615,35 +2633,64 @@ public function run_files_sync(): void
                 $this->state["stage"] = "fetch-skipped";
                 $this->save_state($this->state);
                 $this->run_files_sync_pipeline();
+
+                // Pipeline returns early with partial status if interrupted.
```
adamziel, line 2637: How would we run into this? I'm a bit confused and the comment doesn't tell me what are we guarding against. What is an *interruption* in this comment?

adamziel: Aha, I think I get it now. I'll brush up these comments.

packages/reprint-importer/src/import.php
```diff
@@ -2615,35 +2633,64 @@ public function run_files_sync(): void
...
-
-            $skipped_note = $has_skipped
-                ? " (some files were skipped — re-run with --filter=skipped-earlier to download them)"
-                : "";
+            // Re-run on a completed sync = delta sync (matches the composite
+            // pull's re-pull behavior). Reset run state and fall through to
+            // the fresh-start path below, where $is_delta picks up the local
+            // index and turns the run into re-index → diff → fetch changes.
             $this->audit_log(
-                sprintf("files-pull already complete: %d files indexed%s", $index_size, $skipped_note),
```
adamziel, line 2628: this is valuable audit log information, let's keep it

PR #262, opened 2026-06-15
Add new `—remap` flag for files-pull

WPprodigy:
Tell the importer where to place certain files during `files-pull` via a new `--remap SOURCE TARGET` flag:

It places a source subtree at a chosen path under `--fs-root` instead of nesting it under the source's full absolute path — so instead of content landing at `/srv/htdocs/var/www/html/wp-content/…`, it lands at `/srv/htdocs/wp-content`.

```
--remap :wp-content: :fs-root:/wp-content
--remap :wp-content: /srv/htdocs/wp-content   # same thing as above

# Most-specific source wins
--remap :wp-content: :fs-root:/wp-content --remap :wp-uploads: :fs-root:/media
```

The SOURCE and TARGET are template strings, known `:token:`s are substituted with real paths, and the result must be an absolute path.

  - **SOURCE** — a remote WP-layout token, or a raw absolute remote path:
    - `:wp-content:` `:wp-plugins:` `:wp-mu-plugins:` `:wp-uploads:` — the source site's real component locations, discovered during preflight (so they're correct even when relocated, e.g. uploads under `/user/site-media-files`).
    - `:abspath:` — the source's WP core root, for things outside wp-content (`:abspath:/wp-admin`).
    - a raw absolute path (`/var/www/…`) — pulled verbatim.
  - **TARGET** — a `:fs-root:`-anchored path or a raw absolute path:
    - it must resolve **within `--fs-root`** (otherwise it's an error). `:fs-root:` expands to the `--fs-root` you passed.
 
## Testing

You'll need a source site with the reprint exporter installed.

```
  DOCROOT=/tmp/remap-test  STATE=/tmp/remap-state
  URL="https://YOUR-SITE.com/?reprint-api"  SECRET="<migration secret>"

  rm -rf "$DOCROOT" "$STATE"; mkdir -p "$DOCROOT" "$STATE"

  php importer/import.php preflight  "$URL" --state-dir="$STATE" --fs-root="$DOCROOT" --secret="$SECRET"

  # re-run until it prints "complete" (resumable; exits non-zero while partial)
  php importer/import.php files-pull "$URL" --state-dir="$STATE" --fs-root="$DOCROOT" --secret="$SECRET" \
    --remap :wp-content: :fs-root:/custom/wp-content --no-follow-symlinks
  ```
  
  Then confirm placement:
  - `ls "$DOCROOT/custom/wp-content"` — plugins/ themes/ uploads/ … (landed here via `--remap`)
  - `ls "$DOCROOT"` — `custom/` and `var/` (core falls through, nested under `var/www/html`)

### DIFF
--- a/packages/reprint-importer/src/import.php
+++ b/packages/reprint-importer/src/import.php
@@ -8378,8 +8378,8 @@ class ImportClient
     /**
      * Resolve a --remap argument (source or target) into an absolute path.
      *
-     * Substitutes every known `:token:` (see the token tables in resolve_remap)
-     * with its value, wherever it appears, then trims trailing slashes. The
+     * Substitutes a known leading `:token:` (see the token tables in resolve_remap)
+     * with its value, then trims trailing slashes. The
      * result must be a valid absolute path with no `.`/`..` segments; a relative
      * path or an unknown token (left unsubstituted) fails that check. Referencing
      * a token whose value is unavailable in preflight is a distinct, clear error.
@@ -8391,17 +8391,25 @@ class ImportClient
     {
         $resolved = $raw;
         foreach ($tokens as $name => $value) {
-            if (strpos($resolved, ":{$name}:") === false) {
+            $token = ":{$name}:";
+            $token_offset = strpos($resolved, $token);
+            if ($token_offset === false) {
                 continue;
+            }
+
+            if ($token_offset !== 0 || strpos($resolved, $token, strlen($token)) !== false) {
+                throw new InvalidArgumentException(
+                    "--remap token \"{$token}\" must appear only at the beginning of the path"
+                );
             }
 
             if ($value === null) {
                 throw new InvalidArgumentException(
-                    "Cannot resolve --remap token \":{$name}:\": not available in preflight data. Run preflight first."
+                    "Cannot resolve --remap token \"{$token}\": not available in preflight data. Run preflight first."
                 );
             }
 
-            $resolved = str_replace(":{$name}:", $value, $resolved);
+            $resolved = $value . substr($resolved, strlen($token));
         }
 
         $resolved = rtrim($resolved, "/");
--- a/packages/reprint-importer/src/import.php
+++ b/packages/reprint-importer/src/import.php
@@ -10,6 +10,7 @@
  * - Three-phase import: files, SQL, then file deltas
  */
 
+use function WordPress\Filesystem\wp_join_unix_paths;
 use function WordPress\Reprint\Exporter\assert_valid_path;
 use function WordPress\Reprint\Exporter\normalize_path;
 use function WordPress\Reprint\Exporter\parse_size;
@@ -1132,9 +1133,9 @@ class ImportClient
     private $extra_directory = null;
 
     /**
-     * @var array<int,array{source:string,target:string}> Resolved remap rules:
-     * a full source path → its full local target path (both absolute). Empty =
-     * no remapping (files land nested based on their source).
+     * @var array<string,string> Resolved remap rules: full source path => its
+     * full local target path (both absolute). Empty = no remapping (files land
+     * nested based on their source).
      */
     private $remap_rules = [];
 
@@ -8289,7 +8290,7 @@ class ImportClient
      * here.
      *
      * @param array<int,array{0:string,1:string}> $remap_raw Raw (SRC, TGT) pairs.
-     * @return array<int,array{source:string,target:string}>
+     * @return array<string,string> Source path => target path (both absolute).
      */
     private function resolve_remap(array $remap_raw): array
     {
@@ -8297,31 +8298,23 @@ class ImportClient
         $docroot = rtrim($this->get_filesystem_root_path(), "/");
 
         $rules = [];
-        $explicit_sources = [];
         $wp_content_target = null;
-        foreach ($remap_raw as $pair) {
-            [$src, $tgt] = $pair;
-
-            $wp_path = trim(trim($src), "/");
+        foreach ($remap_raw as [$source, $target]) {
+            $wp_path = trim(trim($source), "/");
             if ($wp_path === "") {
                 throw new InvalidArgumentException("--remap source cannot be empty");
             }
 
             // Accept a target given as a full absolute path that includes the
             // docroot — strip the docroot prefix so it's treated as relative.
-            $tgt = trim($tgt);
-            $under_docroot = self::path_remainder_under(rtrim($tgt, "/"), $docroot);
-            if ($under_docroot !== null) {
-                $tgt = $under_docroot;
-            }
-
-            $tgt_rel = trim($tgt, "/");
+            $target = trim($target);
+            $target_relative = self::path_remainder_under(rtrim($target, "/"), $docroot) ?? $target;
+            $target_relative = trim($target_relative, "/");
 
             $source = $this->resolve_source_path($wp_path, $source_paths);
-            $target = $tgt_rel === "" ? $docroot : $docroot . "/" . $tgt_rel;
-            $rules[] = ["source" => $source, "target" => $target];
+            $target = wp_join_unix_paths($docroot, $target_relative);
+            $rules[$source] = $target;
 
-            $explicit_sources[$source] = true;
             if ($wp_path === "wp-content") {
                 $wp_content_target = $target;
             }
@@ -8331,9 +8324,9 @@ class ImportClient
         if ($wp_content_target !== null) {
             foreach (["plugins", "mu-plugins", "uploads"] as $name) {
                 $source = $source_paths[$name];
-                $detached = self::path_remainder_under($source, $source_paths["content"]) === null;
-                if ($detached && !isset($explicit_sources[$source])) {
-                    $rules[] = ["source" => $source, "target" => $wp_content_target . "/" . $name];
+                $is_outside_content_dir = self::path_remainder_under($source, $source_paths["content"]) === null;
+                if ($is_outside_content_dir && !isset($rules[$source])) {
+                    $rules[$source] = wp_join_unix_paths($wp_content_target, $name);
                 }
             }
         }
@@ -8345,8 +8338,9 @@ class ImportClient
      * The source site's real component locations from preflight data, as
      * name => absolute path (abspath, content, plugins, mu-plugins, uploads).
      * Each wp-content component falls back to its conventional spot under
-     * content_dir; one resolving outside content_dir is "detached". abspath
-     * anchors any path outside wp-content and may be null if undetermined.
+     * content_dir; one resolving outside content_dir is relocated and routed
+     * separately. abspath anchors any path outside wp-content and may be null
+     * if undetermined.
      */
     private function wp_component_source_paths(): array
     {
@@ -8396,7 +8390,7 @@ class ImportClient
         foreach ($bases as $prefix => $base) {
             $rest = self::path_remainder_under($wp_path, $prefix);
             if ($rest !== null) {
-                return $base . $rest;
+                return wp_join_unix_paths($base, $rest);
             }
         }
         // Anything outside wp-content (wp-admin, core, a root file, ...) is
@@ -8407,7 +8401,7 @@ class ImportClient
             );
         }
 
-        return $source_paths["abspath"] . "/" . $wp_path;
+        return wp_join_unix_paths($source_paths["abspath"], $wp_path);
     }
 
     /**
@@ -8422,13 +8416,13 @@ class ImportClient
     private function remap_source_path_to_target(string $source_path): ?string
     {
         $best = null;
-        $best_source_len = -1;
+        $best_source_length = -1;
 
-        foreach ($this->remap_rules as $rule) {
-            $rest = self::path_remainder_under($source_path, $rule["source"]);
-            if ($rest !== null && strlen($rule["source"]) > $best_source_len) {
-                $best = $rule["target"] . $rest;
-                $best_source_len = strlen($rule["source"]);
+        foreach ($this->remap_rules as $source => $target) {
+            $rest = self::path_remainder_under($source_path, $source);
+            if ($rest !== null && strlen($source) > $best_source_length) {
+                $best = wp_join_unix_paths($target, $rest);
+                $best_source_length = strlen($source);
             }
         }
 
@@ -9300,8 +9294,10 @@ class ImportClient
         // Ensure every --remap source is enumerated — including a detached
         // component (relocated uploads/plugins dir) that lives outside the
         // WordPress roots and so wouldn't be discovered by traversal alone.
-        foreach ($this->remap_rules as $i => $rule) {
-            $extra_paths["remap_src_{$i}"] = $rule["source"];
+        $remap_index = 0;
+        foreach (array_keys($this->remap_rules) as $source) {
+            $extra_paths["remap_source_{$remap_index}"] = $source;
+            $remap_index++;
         }
 
         // auto_prepend_file / auto_append_file may point to directories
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: packages/reprint-importer/src/import.php [hunk 1]
```diff
@@ -8269,6 +8282,158 @@ private function get_filesystem_root_path(): string
...
+
+            $explicit_sources[$source] = true;
+            if ($wp_path === "wp-content") {
+                $wp_content_target = $target;
+            }
+        }
+
+        // Pass 2: Do a whole tree remap for wp-content's components if needed.
+        if ($wp_content_target !== null) {
+            foreach (["plugins", "mu-plugins", "uploads"] as $name) {
```
atanas-dev: How about `themes/` and `languages/`? Plugins occasionally add their own directories in `wp-content/`, too (and files!).

Instead of mapping its contents, can we map `wp-content` itself?

atanas-dev (commented):
Makes sense. I haven't tested this; I just gave it a once-over.

packages/reprint-importer/src/import.php
WPprodigy: It is mapping wp-content itself. So the additional things inside wp-content will automatically come over as well (including themes and languages). 

These three are just special because they can become "detached" and located somewhere outside of wp-content. So this pass 2 mapping is using the configs it got during preflight, asking the source WP site where it's uploads_dir is, etc.

So when the uploads_dir that we see in preflight is NOT underneath wp-content, then what I'm doing in the above is explicitely remapping it to be underneath wp-content as somebody would expect when asking to remap wp-content to somewhere new.

So`/var/www/html/special-uploads-dir` => `/srv/htdocs/wp-content/uploads` becomes a remapping.

WPprodigy: I've split the behavior in a method in an upcoming PR (`detached_component_sources()`) so it can be re-used with the `--only` command as well. So the code should read a bit clearer as a result.

packages/reprint-importer/src/import.php
```diff
@@ -8269,6 +8282,158 @@ private function get_filesystem_root_path(): string
...
+    private function resolve_remap(array $remap_raw): array
+    {
+        $source_paths = $this->wp_component_source_paths();
+        $docroot = rtrim($this->get_filesystem_root_path(), "/");
+
+        $rules = [];
+        $explicit_sources = [];
+        $wp_content_target = null;
+        foreach ($remap_raw as $pair) {
+            [$src, $tgt] = $pair;
```
adamziel: Let's use words in variable names everywhere in this PR, they're easier to process for humans than abbreviations: 

```suggestion
            [$source, $target] = $pair;
```

packages/reprint-importer/src/import.php
```diff
@@ -8269,6 +8282,158 @@ private function get_filesystem_root_path(): string
...
+
+            $wp_path = trim(trim($src), "/");
```
adamziel: Trimming `/` allows for an user error here, as in `--remap /home/adam/logs ./logs`. Let's detect it and point it out:

```suggestion
            if(substr($source, 0, 1) === '/') {
                throw new InvalidArgumentException("--remap source must be a <how do we explain it?>");
            }
            $wp_path = rtrim(trim($source), "/");
```

adamziel: Also, I wouldn't trim whitespaces from `$source`:

```
> echo "I am a valid file" > "file "
> cat "file "
I am a valid file

> cat file
cat: file: No such file or directory
```

packages/reprint-importer/src/import.php
```diff
@@ -8269,6 +8282,158 @@ private function get_filesystem_root_path(): string
...
+            if ($wp_path === "") {
+                throw new InvalidArgumentException("--remap source cannot be empty");
+            }
+
+            // Accept a target given as a full absolute path that includes the
+            // docroot — strip the docroot prefix so it's treated as relative.
+            $tgt = trim($tgt);
```
adamziel: Ditto for the target path. It may not matter as much in WP Cloud, but it may matter in the general case.

packages/reprint-importer/src/import.php
```diff
@@ -8277,14 +8442,43 @@ private function get_filesystem_root_path(): string
...
+        }
         return $this->get_filesystem_root_path() . $path;
     }
 
+    /**
+     * Returns the remainder of $path underneath $prefix,
+     * empty string if $path === $prefix,
+     * or null if $path is not under $prefix.
+     */
+    private static function path_remainder_under(string $path, string $prefix): ?string
```
adamziel, line 8527: I've struggled to find a good name for that helper in other projects. Thank you!

adamziel: It misses `$path = '/home/adam` and `$prefix = '/home/adam/`

packages/reprint-importer/src/import.php
```diff
@@ -8269,6 +8282,158 @@ private function get_filesystem_root_path(): string
...
+
+            $wp_path = trim(trim($src), "/");
+            if ($wp_path === "") {
+                throw new InvalidArgumentException("--remap source cannot be empty");
+            }
+
+            // Accept a target given as a full absolute path that includes the
+            // docroot — strip the docroot prefix so it's treated as relative.
+            $tgt = trim($tgt);
+            $under_docroot = self::path_remainder_under(rtrim($tgt, "/"), $docroot);
```
adamziel: It would read so much nicer as

```suggestion
            $under_docroot = self::path_remainder_under($target, $docroot);
```

adamziel: Or even

```suggestion
            $target = self::path_remainder_under($target, $docroot) ?? $target;
```

packages/reprint-importer/src/import.php
```diff
@@ -8269,6 +8282,158 @@ private function get_filesystem_root_path(): string
...
+            if ($under_docroot !== null) {
+                $tgt = $under_docroot;
+            }
+
+            $tgt_rel = trim($tgt, "/");
+
+            $source = $this->resolve_source_path($wp_path, $source_paths);
+            $target = $tgt_rel === "" ? $docroot : $docroot . "/" . $tgt_rel;
```
adamziel: We could bring some utilities from php-toolkit, e.g. [wp_join_unix_paths()](https://github.com/WordPress/php-toolkit/blob/947734a2823675292580cb6252e5e158e5521e61/components/Filesystem/functions.php#L152):


```suggestion
            $target = wp_join_unix_paths( $docroot, $target_relative );
```

packages/reprint-importer/src/import.php
```diff
@@ -8269,6 +8282,158 @@ private function get_filesystem_root_path(): string
...
+            $rules[] = ["source" => $source, "target" => $target];
+
+            $explicit_sources[$source] = true;
```
adamziel: Do we need two arrays? Versus `$rules[$source] = $target`? The order of insertions is preserved.

packages/reprint-importer/src/import.php
```diff
@@ -8269,6 +8282,158 @@ private function get_filesystem_root_path(): string
...
+            if ($wp_path === "wp-content") {
+                $wp_content_target = $target;
+            }
+        }
+
+        // Pass 2: Do a whole tree remap for wp-content's components if needed.
+        if ($wp_content_target !== null) {
+            foreach (["plugins", "mu-plugins", "uploads"] as $name) {
+                $source = $source_paths[$name];
```
adamziel: Any chance either of these directories is missing on the source site?

packages/reprint-importer/src/import.php
```diff
@@ -8269,6 +8282,158 @@ private function get_filesystem_root_path(): string
...
+                $detached = self::path_remainder_under($source, $source_paths["content"]) === null;
```
adamziel: Could we call it `$is_outside_of_document_root`? That wouldn't require introducing a new term.

packages/reprint-importer/src/import.php
```diff
@@ -8269,6 +8282,158 @@ private function get_filesystem_root_path(): string
...
+     * Each wp-content component falls back to its conventional spot under
+     * content_dir; one resolving outside content_dir is "detached". abspath
+     * anchors any path outside wp-content and may be null if undetermined.
+     */
+    private function wp_component_source_paths(): array
+    {
+        $preflight = $this->state["preflight"]["data"] ?? [];
+        $paths = $preflight["database"]["wp"]["paths_urls"] ?? [];
+
+        $content_dir = $this->flatten_clean_path($paths["content_dir"] ?? null);
```
adamziel: `content_dir` is the remote site's `WP_CONTENT_DIR` and could theoretically be a symlink. I don't think this impacts us negatively here, but I just want to voice it.

packages/reprint-importer/src/import.php
```diff
@@ -8269,6 +8282,158 @@ private function get_filesystem_root_path(): string
...
+
+        $abspath = $this->flatten_clean_path($paths["abspath"] ?? null);
+        if ($abspath === null) {
+            $abspath = $this->flatten_clean_path($preflight["wp_detect"]["roots"][0]["path"] ?? null);
+        }
+
+        return [
+            "abspath" => $abspath,
+            "content" => $content_dir,
+            "plugins" => $this->flatten_clean_path($paths["plugins_dir"] ?? null) ?? $content_dir . "/plugins",
```
adamziel: `$content_dir . "/plugins"` assumes that directory exists on the remote site. Can we confirm that in the index? Ditto for the other lines below.

packages/reprint-importer/src/import.php
```diff
@@ -8269,6 +8282,158 @@ private function get_filesystem_root_path(): string
...
+        }
+        // Anything outside wp-content (wp-admin, core, a root file, ...) is
+        // taken relative to ABSPATH.
+        if ($source_paths["abspath"] === null) {
+            throw new InvalidArgumentException(
+                "Cannot resolve --remap {$wp_path}: preflight has no ABSPATH. Run preflight first.",
+            );
+        }
+
+        return $source_paths["abspath"] . "/" . $wp_path;
```
adamziel: `wp_join_unix_paths()` could be useful for always handling slashes correctly in these.

packages/reprint-importer/src/import.php
```diff
@@ -8269,6 +8282,158 @@ private function get_filesystem_root_path(): string
...
+     */
+    private function remap_source_path_to_target(string $source_path): ?string
+    {
+        $best = null;
+        $best_source_len = -1;
+
+        foreach ($this->remap_rules as $rule) {
+            $rest = self::path_remainder_under($source_path, $rule["source"]);
+            if ($rest !== null && strlen($rule["source"]) > $best_source_len) {
+                $best = $rule["target"] . $rest;
```
adamziel: another good place for `wp_join_unix_paths()`

packages/reprint-importer/src/import.php
```diff
@@ -9103,6 +9297,13 @@ private function get_export_directories(): array
             $extra_paths["extra_directory"] = rtrim($this->extra_directory, "/");
         }
 
+        // Ensure every --remap source is enumerated — including a detached
+        // component (relocated uploads/plugins dir) that lives outside the
```
adamziel, line 9362: Is it actually possible to reference a detached component that wouldn't otherwise be visited? Would we need it if there was no symlink to it within the document root? I suppose a custom nginx config could serve uploads from a completely detached location so maybe yes.

Still, I keep getting confused here. `--remap wp-content/uploads ./uploads` is effectively an identifier of the uploads directory that is mapped to the relevant path on the source site. It could actually live under `/user/site-media-files` which is what throws me off. What if we used template strings instead of path-like identifiers? e.g. `--remap :wp-uploads: ./uploads`? Or `--remap :wp-plugins:jetpack ./uploads` where the `:wp-plugins:` would be substituted for the remote path?

WPprodigy: I like this! Switching to template strings.

packages/reprint-importer/src/import.php
WPprodigy: So if the preflight hasn't run yet and found a wp-content dir, it'll error out earlier on in wp_component_source_paths(): `Cannot resolve --remap: preflight has no content_dir. Run preflight first.`

With preflight present, a component path that doesn't exist on the source is basically a no-op anyway. The exporter enumerates nothing under it, so the rule never matches a file.

packages/reprint-importer/src/import.php
WPprodigy: Is possible I suppose, but we can only give it best effort. We consult the preflight for information on where they exist. If they don't exist, then there's just not going to be any files being transferred under these paths, so the remapping won't be doing anything anyway.

adamziel: That works! I mostly want to make sure it won't derail the whole procedure. 👍

packages/reprint-importer/src/import.php
adamziel: Ah, cool! That would make a great inline comment for when we read this code in the future.

packages/reprint-importer/src/import.php
```diff
@@ -8269,6 +8283,157 @@ private function get_filesystem_root_path(): string
...
 
+    /**
+     * Build the remap rules from the raw remap pairs + preflight data.
+     *
+     * Each argument is a template string of `:token:` substitutions and/or a raw absolute path.
+     * Source arguments resolve against the remote site's component locations.
+     * Target arguments resolve under --fs-root and must stay within it.
+     * Each rule is a full source path => full local target path (both absolute).
+     *
+     * @param array<int,array{0:string,1:string}> $remap_raw Raw (SRC, TGT) pairs.
```
adamziel: Non-blocking nitpick: 

```suggestion
     * @param array<int,array{0:string,1:string}> $remap_raw Raw (SOURCE, TARGET) pairs.
```

packages/reprint-importer/src/import.php
```diff
@@ -8269,6 +8283,157 @@ private function get_filesystem_root_path(): string
...
+     * @return array<string,string> Source path => target path (both absolute).
+     */
+    private function resolve_remap(array $remap_raw): array
+    {
+        $source_paths = $this->wp_component_source_paths();
+        $fs_root = rtrim($this->get_filesystem_root_path(), "/");
+
+        $source_tokens = [
```
adamziel: Should we just do this:

```php
        $source_tokens = $this->wp_component_source_paths();
```

And return the correctly keyed array from that method so that we don't need to re-map it?

packages/reprint-importer/src/import.php
```diff
@@ -8269,6 +8283,157 @@ private function get_filesystem_root_path(): string
...
+     * a token whose value is unavailable in preflight is a distinct, clear error.
+     *
+     * @param string $raw The raw remap argument.
+     * @param array<string,string|null> $tokens Token name => value (null = unavailable).
+     */
+    private function resolve_remap_path(string $raw, array $tokens): string
+    {
+        $resolved = $raw;
+        foreach ($tokens as $name => $value) {
+            if (strpos($resolved, ":{$name}:") === false) {
```
adamziel: The token may only be the very first thing in the raw string, right? We might want to check if the strpos is 0

packages/reprint-importer/src/import.php
adamziel: I took a stab at it, feel free to revert.

packages/reprint-importer/src/import.php
adamziel: I took a stab at it, feel free to revert.

packages/reprint-importer/src/import.php
adamziel: I took a stab at it, feel free to revert.

packages/reprint-importer/src/import.php
WPprodigy: Sure, works for now. I end up putting this in a helper anyway though during the --only PR since the token system ends up being re-used.

packages/reprint-importer/src/import.php
WPprodigy: Yeah, all the tokens right now really only make sense at the start of the string.

PR #266, opened 2026-06-18
Strip ephemeral _edit_lock postmeta on db-apply

epeicher:
## What it does

A site pulled with reprint no longer shows a phantom "**X is currently editing**" badge in wp-admin. `db-apply` now strips the remote site's editor locks before marking the import complete.

## Rationale

A WordPress SQL dump captures `_edit_lock` postmeta verbatim. `_edit_lock` is ephemeral runtime state, recording that an editor had a given post open at the moment of the dump (its value is `timestamp:user_id`). On the imported copy the lock is meaningless, the remote editing session has nothing to do with the local site, but WordPress still renders it as a "currently editing" badge in the posts list for up to ~150 seconds after the captured timestamp whenever the pull ran while a post was open remotely.

## Implementation

Added `strip_ephemeral_edit_locks(PDO $pdo)`, called at the end of `run_db_apply` while the target PDO is still open (right after host-plugin deactivation, before the state is marked complete):

```php
$postmeta_table = '`' . str_replace('`', '``', $table_prefix . 'postmeta') . '`';
$deleted = $pdo->exec("DELETE FROM {$postmeta_table} WHERE meta_key = '_edit_lock'");
// The dump runs with AUTOCOMMIT=0 and leaves it off, so the DELETE needs an explicit COMMIT.
$pdo->exec('COMMIT');
```

- The table prefix comes from preflight data (defaults to `wp_`); the name is backtick-quoted with `` ` `` doubling to prevent injection from a crafted prefix.
- Only `_edit_lock` rows are removed; all other postmeta is untouched. WordPress regenerates the lock locally on the next edit, and core's own WXR exporter skips it too, so nothing is lost.
- A dump without a postmeta table is tolerated: the whole operation is wrapped in `try/catch` and logged to the audit log, never fatal.

## Testing instructions

The easiest way to reproduce the issue is to use Studio code:

- Open the remote site, add a Post and publish it, but don't close the window.
- Navigate to Studio repo root and run:
```
export STUDIO_ENABLE_PULL_REPRINT=true && node ~/github/studio/apps/cli/dist/cli/main.mjs pull-reprint --url <your_wpcom_site>
```
- Reply `Y` to create a new Studio site
- Once the site is created, you can open the URL for the `WP Admin` returned by the Reprint CLI
- Observe that the created post is there with a `Currently being edited` message 
CleanShot 2026-06-18 at 11 20 10@2x

- Apply the changes on this branch, run `composer build:phar`, copy that to Studio repo by running `cp reprint.phar <studio_repo>/wp-files/reprint`
- On Studio, run `npm run cli:build` so the phar file is copied to the dist folder
- Delete the site to start from scratch and repeat the first three steps
- Observe that the created post does not have the `Currently being edited` message

### DIFF
--- a/packages/reprint-importer/src/import.php
+++ b/packages/reprint-importer/src/import.php
@@ -5402,6 +5402,11 @@ class ImportClient
                     $this->audit_log("DB-APPLY | deactivated plugin {$basename} (path-incompatible siteurl)");
                 }
 
+                // Drop editor locks captured from the remote. On the imported
+                // copy they only produce phantom "X is currently editing"
+                // badges.
+                $this->strip_ephemeral_edit_locks($pdo);
+
                 // Mark complete
                 $this->state["apply"]["statements_executed"] = $statements_executed;
                 $this->state["apply"]["bytes_read"] = $seek_offset + $query_stream->get_bytes_consumed();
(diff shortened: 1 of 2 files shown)


adamziel:
solved in https://github.com/WordPress/reprint/pull/271

PR #284, opened 2026-06-30
Add support for --env-mysql-* flags for secure credential passing

### LINKED ISSUE #24: Support --env-db-password=DB_PASSWORD
Including database credentials in the CLI command is risky in that anyone with `ps` access may read them. Let's support `import.php` CLI flags to point the script to specific ENV variables holding the database username, host, password, etc. They would be alternatives to `--db-*` flags, so there would be `--db-password=password` and an alternative flag `--env-db-password=DB_PASSWORD` that would source the password from the `DB_PASSWORD` env variable.

TheVaibhaw:
Implements feature requested in issue #24:
- Add --env-mysql-password=VAR_NAME support
- Add --env-mysql-user=VAR_NAME support
- Add --env-mysql-host=VAR_NAME support
- Add --env-mysql-database=VAR_NAME support

These flags allow credentials to be sourced from environment variables specified by the user, avoiding password visibility in 'ps aux' output.

Maintains backward compatibility with existing --mysql-* flags. Falls back to hardcoded MYSQL_PASSWORD env var if no flag specified.

Changes:
- Add CLI option definitions for all 4 new flags
- Update credential parsing logic to check env flags first
- Only persist non-password credentials in state

Fixes #24

### DIFF
--- a/packages/reprint-importer/src/import.php
+++ b/packages/reprint-importer/src/import.php
@@ -1681,7 +1681,13 @@ class ImportClient
         }
 
         // MySQL connection parameters for --sql-output=mysql.
-        if (isset($options["mysql_host"])) {
+        if (isset($options["env_mysql_host"])) {
+            $env_var_name = $options["env_mysql_host"];
+            $this->mysql_host = getenv($env_var_name) ?: '';
+            if (!empty($this->mysql_host)) {
+                $this->state["mysql_host"] = $this->mysql_host;
+            }
+        } elseif (isset($options["mysql_host"])) {
             $this->mysql_host = $options["mysql_host"];
             $this->state["mysql_host"] = $this->mysql_host;
         } elseif (isset($this->state["mysql_host"])) {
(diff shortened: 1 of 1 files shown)


packages/reprint-importer/src/import.php
```diff
@@ -1681,7 +1681,13 @@ public function run(array $options = []): void
         }
 
         // MySQL connection parameters for --sql-output=mysql.
-        if (isset($options["mysql_host"])) {
+        if (isset($options["env_mysql_host"])) {
+            $env_var_name = $options["env_mysql_host"];
+            $this->mysql_host = getenv($env_var_name) ?: '';
+            if (!empty($this->mysql_host)) {
```
adamziel, line 1687: I'd remove this check. If we're telling the tool to use a specific ENV, let it use specific ENV. We're not checking for non-empty values in other branches. Other than that, this PR looks good!

PR #292, opened 2026-07-01
Better prior-op understanding for `files-sync --filter=skipped-earlier` calls

fredrikekelund:
### Background

The `pull-reprint` command in Studio CLI runs Reprint in the "two-phased pull mode", which means a first pass with `pull --filter=essential-files` and a second pass with `files-sync --filter=skipped-earlier`. The entire point of this approach is so that we can start the local WordPress server in between the first and second pass, so the user can get to work sooner. 

This is a nice touch from the user perspective, but the Reprint implementation has flaws related to state handling in the `files-sync --filter=skipped-earlier` pass that are currently taped over in the Studio CLI `pull-reprint` implementation This is the first PR of at least two where my ultimate goal is to update Reprint in a way that lets us drop the logic in the Studio CLI `pull-reprint` command that mutates Reprint's internal state (see [this](https://github.com/Automattic/studio/blob/65f108b39cfd553f0916b7e79fe41bc9d5cab03b/apps/cli/commands/pull-reprint.ts#L821-L856)). 

### Changes in this PR

This PR fixes two issues.

**Firstly**, a `files-sync --filter=skipped-earlier` run that's interrupted leaves `filter: skipped-earlier` and `active_resumable_command.completion_state: partial` in the disk state. The next `pull --filter=essential-files` run then throws:

```
Cannot change --filter from 'skipped-earlier' to 'essential-files' while a sync is in progress.
```

This is because Reprint thinks the previously interrupted `files-sync --filter=skipped-earlier` operation must run to completion before it's safe to run a `--filter=essential-files` pull, but that's not actually true. It is safe to restart the pull in this scenario. Claude added the code that makes Reprint do this.

**Secondly**, there's [logic](https://github.com/Automattic/studio/blob/65f108b39cfd553f0916b7e79fe41bc9d5cab03b/apps/cli/commands/pull-reprint.ts#L821-L856) in the `pull-reprint` command in Studio CLI (the same as I was referencing earlier) that tapes over an issue with running `files-sync --filter=skipped-earlier` after `pull --filter=essential-files`.

If I remove that logic and run a pull, I get this error:

```
--filter=skipped-earlier was requested but there is no completed sync with skipped files.
```

This happens despite the initial pull completing successfully and leaving a skipped list. That's because `run_files_sync()` doesn't properly consider the actual end state that `pull --filter=essential-files` leaves behind. The logic in the Studio CLI `pull-reprint` command clearly illustrates the problem. Claude implemented a fix for this.

Ultimately, we should probably update the `pull-files` command to handle `--filter=skipped-earlier` and move the status rigmarole into `run_pipeline()`. I'll leave that for another PR, though.

### Testing instructions

TBD

### DIFF
--- a/packages/reprint-importer/src/import.php
+++ b/packages/reprint-importer/src/import.php
@@ -1660,7 +1660,9 @@ class ImportClient
         //
         // Changing the filter mid-flight is not allowed.  The user must either
         // start fresh (--abort) or finish the current sync before switching.
-        // The one valid transition is: essential-files (complete) → skipped-earlier.
+        // Exceptions: essential-files may advance to skipped-earlier once
+        // complete, and skipped-earlier (a terminal tail-fetch) may always give
+        // way to a fresh full sync.
         if (isset($options["filter"])) {
             $next = $options["filter"];
             if (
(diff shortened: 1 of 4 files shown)


adamziel (approved):
@fredrikekelund this works and makes sense, thank you! I'm happy for this to land, and also I just had this thought: What if we dropped the concept of a `filter` entirely, and replaced it with `--only :wp-content:/uploads` and `--except :wp-content:/uploads`? `--only` is already implemented, `--except` shouldn't be difficult to add, and that's what `--filter` amounts to anyway. The only difference is `filter` remembers what it skipped over and reuses that list without consulting with the remote site again.

PR #328, opened 2026-07-08
Add --follow-symlinks=DIR, consolidating out-of-scope symlinks

WPprodigy:
Migrations (especially scoped ones with `--only :wp-content:`) don't have a great option for symlinks that point outside the pulled scope. Bare `--follow-symlinks` pulls the content but mirrors it at its raw source paths, scattered across the docroot. `--no-follow-symlinks` just drops it, leaving dangling links.

This PR adds a destination form: `--follow-symlinks=:fs-root:/.symlinks-bundle`. Escaping targets get followed and consolidated into one private in-tree directory, nested by source path, and their symlinks are repointed there. A scoped wp-content pull ends up self-contained: no dangling links, no junk in the docroot.

**Placement**: remap rules first, then the bundle location if set, then the default fs-root mirror. Anything inside a remapped wp-content lands in the served tree, so the bundle only receives genuinely-external content. Multiple links to one target share a single copy. Relative and absolute targets are handled the same. Bare `--follow-symlinks` is equivalent to `--follow-symlinks=:fs-root:` (the unchanged default), and the value uses the `--remap` target grammar — it must stay within `--fs-root`.

**Resume guard**: the bundle directory is fingerprinted on the first run and checked on every run after. Changing (or dropping) it errors instead of silently splitting placement; `--abort` resets it. Setting a bundle directory also persists `follow_symlinks=true`, so a flag-less delta can't index with follow off and delete previously-followed content.

**Behavior changes** (both previously produced missing content and dangling links):
1. Under a narrow `--only`, followed targets outside the `--only` scope are now pulled. They used to be skipped as "already covered" because they sat under a preflight root, even though the scoped pull never actually downloaded them.
2. Symlinks in the middle of a followed path (e.g. `/srv/wordpress → /wordpress`) are now repointed at the locally-mirrored content, like every other symlink. They used to be skipped whenever their absolute target pointed outside `--fs-root`, even when the content itself had been downloaded.

## Testing

The new `SymlinkBundleTest` (unit) + `import-52-symlink-bundle` (e2e, broad + narrow `--only`) cover lots of cases here.

For manual testing though, the setup would be rather intense. You'll need something like a symlinked plugin that points outside wp-content. Then run `--only :wp-content: --follow-symlinks=:fs-root:/.symlinks-bundle` and ensure the plugin's source ended up in the new bundle location.

### DIFF
--- a/packages/reprint-importer/src/import.php
+++ b/packages/reprint-importer/src/import.php
@@ -1090,6 +1090,18 @@ class ImportClient
     private $follow_symlinks = true;
 
     /**
+     * @var string|null Destination directory for escaping (out-of-scope) symlink targets,
+     * nested by the source path. null keeps the default: each symlink target is
+     * mirrored at its own source path underneath --fs-root.
+     *
+     * Usage: --follow-symlinks=<dir>
+     */
+    private $symlink_bundle_directory = null;
+
+    /** @var array|null Cached result of get_export_directories(). */
+    private $export_directories_cache = null;
+
+    /**
      * @var bool When true, ask the server to ship the default-skipped
      * generated content (wp-content/cache, .git, node_modules, etc.).
      *
@@ -1642,6 +1654,13 @@ class ImportClient
             $this->save_state($this->state);
         } elseif (isset($this->import_state()->follow_symlinks)) {
             $this->follow_symlinks = $this->import_state()->follow_symlinks;
+        }
+
+        if (isset($options["symlink_bundle_directory"])) {
+            $this->symlink_bundle_directory = $this->resolve_symlink_bundle_directory($options["symlink_bundle_directory"]);
+            $this->follow_symlinks = true;
+            $this->import_state()->follow_symlinks = true;
+            $this->save_state($this->state);
         }
 
         // Persist fs_root_nonempty_behavior in state so it survives across invocations.
@@ -2694,6 +2713,7 @@ class ImportClient
 
         $this->recover_index_updates();
         $this->assert_files_pull_only_unchanged_while_resuming($has_progress);
+        $this->assert_symlink_bundle_directory_unchanged();
 
         // Already completed.
         if ($current_status === "complete") {
@@ -3237,7 +3257,10 @@ class ImportClient
      */
     private function discover_symlink_targets(): void
     {
-        $roots = $this->get_root_directories_from_preflight();
+        // Seed "already covered" from the dirs actually enumerated this run (the
+        // --only prefixes when scoped), not the full preflight roots — otherwise a
+        // narrow --only skips a target under a root but outside its scope.
+        $roots = $this->get_export_directories();
 
         // Collect all indexed directory real paths for containment checks
         $visited = [];
@@ -3492,6 +3515,11 @@ class ImportClient
                 );
                 continue;
             }
+
+            // Repoint through the same seam regular symlink chunks use, so the
+            // link targets wherever the content actually landed (fs-root,
+            // remapped, or bundled) instead of the raw source spelling.
+            $target = $this->map_symlink_target_for_local_mirror($path, $local_path, $target);
 
             // Already correct — skip
             if (is_link($local_path) && readlink($local_path) === $target) {
@@ -8317,7 +8345,11 @@ class ImportClient
     }
 
     /**
-     * Map an absolute remote symlink target to the local fs-root mirror when possible.
+     * Map a remote symlink target to the local fs-root mirror when possible.
+     *
+     * Handles both absolute and relative targets (relative ones are resolved
+     * against the symlink's source directory). In-scope and non-followed targets
+     * keep their original spelling.
      *
      * Example:
      *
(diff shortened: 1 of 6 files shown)


packages/reprint-importer/src/import.php
```diff
@@ -1644,6 +1656,13 @@ public function run(array $options = []): void
             $this->follow_symlinks = $this->import_state()->follow_symlinks;
         }
 
+        if (isset($options["symlink_bundle_directory"])) {
+            $this->symlink_bundle_directory = $this->resolve_symlink_bundle_directory($options["symlink_bundle_directory"]);
+            $this->follow_symlinks = true;
+            $this->import_state()->follow_symlinks = true;
```
adamziel, line 1662: Unrelated to this PR: Maybe we can reduce that to a single flag 🤔

packages/reprint-importer/src/import.php
```diff
@@ -8384,33 +8378,29 @@ private function assert_symlink_target_within_root(
...
-            !$this->remote_index_contains_path_prefix($normalized_target)
+            !$this->remote_index_contains_path_prefix($source_target)
         ) {
             return $target;
         }
 
-        $mapped_absolute = $root . $normalized_target;
+        // Repoint to wherever the target's content is actually placed,
+        // the same seam file chunks use, so that the link never dangles.
+        $mapped_absolute = $this->remote_path_to_local_path_within_import_root($source_target);
```
adamziel, line 8403: Do we preserve symlinks structures like this?

```
/wordpress/wp-content/plugins/myhost-cache -> /plugins/myhost-cache
/plugins/myhost-cache
/plugins/myhost-cache/uploads -> /wordpress/wp-content/uploads
```

If not: I wouldn't expect them to come up often and we also don't have to block this PR on that, but I'd acknowledge it as an issue on GitHub and maybe a TODO comment in the codebase.

packages/reprint-importer/src/import.php
```diff
@@ -8881,6 +8927,13 @@ private function remote_path_to_local_path_within_import_root(
                 return $target;
             }
         }
+
+        // A followed target outside the original export scope goes underneath the bundle directory.
+        if ($this->symlink_bundle_directory !== null
+            && !$this->path_is_within_original_export_scope($path)) {
```
adamziel, line 8934: This seems off. It assumes `$path` is a result of following a symlink. While I think it's true today, it's an implicit assumption that may break at any point. Is this code path in here just to support this line in `map_symlink_target...` method above?

```php
$mapped_absolute = $this->remote_path_to_local_path_within_import_root($source_target);
```

If yes, let's just inline this logic into that other method to make the assumption strong and explicit.

packages/reprint-importer/src/import.php
```diff
@@ -9707,16 +9776,24 @@ private function get_root_directories_from_preflight(): array
      */
     private function get_export_directories(): array
     {
+        // Memoized: The inputs (pull_only, remap, preflight) are all set before
+        // the first caller and never change mid-run, so cache on first use.
+        if ($this->export_directories_cache !== null) {
+            return $this->export_directories_cache;
```
adamziel, line 9783: Does it solve some performance problem you've stumbled upon?

packages/reprint-importer/src/import.php
WPprodigy: Yeah, could definitely merge these together a bit.

I've actually go a WIP draft that refactors and simplifies how we're doing per-flag resume guards & carrying over the state between runs, using a single identity hash for a run up front.

Could see folding this into that.

packages/reprint-importer/src/import.php
WPprodigy: - The bundle scope check consults this per placed file now, while previously it was only per index request.
- It was also double-logging the `DIRECTORY AUTO-DETECT` audit log on ever call when not using `--only`.

packages/reprint-importer/src/import.php
WPprodigy: Yes, those are preserved. I tested this specific structure in an e2e test to double confirm as well.

Symlink targets get repointed using the same placement logic as regular files, so the back-link ends up pointing at wherever wp-content/uploads actually landed locally, rather than at its raw source path.

packages/reprint-importer/src/import.php
WPprodigy: It's not just here for repointing the symlink, it also dictates where the actual folder/files land. If we moved to the other method, the symlinks would point to the bundle location but  the actual content would end up landing scattered at ` fs-root/<path> ` still, defeating the whole purpose.

The assumption is that "content outside the export scope lands under the bundle directory." And, at least right now, following symlinks is the only source of out-of-scope content.

adamziel: OK I think that's fine. Let's just call that explicitly in the comment. And then let's merge 🚀

packages/reprint-importer/src/import.php
adamziel: right, but does caching the value has a measurable effect? as in, is the additional gotcha "this may return stale data" worth it because of the added speed?

PR #335, opened 2026-07-09
Support PHP 7.2 in Reprint exporter

fredrikekelund:
### Background

Our initial strategy for shipping the Reprint exporter to Pressable sites is to include it in the stock Jetpack plugin. The Reprint exporter already lives in wpcomsh (which is part of Jetpack), so the tactic is:

1. Duplicate that implementation in the regular Jetpack plugin, gating the Reprint exporter feature to Pressable and WoA sites.
2. After deploying that change, update the Reprint client to use the new Jetpack API URLs to interact with the Reprint exporter.
3. Remove the wpcomsh Reprint exporter implementation from the Jetpack repo.

One obvious concern here is, "Is it really good UX to deploy this through Jetpack?" The answer is that we are doing this to support Studio sync, which already depends on Jetpack for Pressable sites to work, and Jetpack also gives us auth. Long-term, it is very likely there'll be a standalone Reprint exporter plugin, just not to begin with.

### What is this PR?

The standard Jetpack plugin supports PHP 7.2. To be able to include the Reprint exporter there, we need to update the Reprint exporter code to work with PHP 7.2 This PR does just that.

The client is not affected. You can still use newer PHP syntax and features there.

### Testing instructions 

CI should pass

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -117,9 +117,11 @@ jobs:
           SECRET="test-secret-basic"
           BASE="http://127.0.0.1:8081/?reprint-api"
           DIR="/srv/e2e-sites/basic"
+          STATE_DIR="/tmp/e2e-smoke-state"
+          FS_ROOT="/tmp/e2e-smoke-fs"
 
           echo "=== Importer preflight ==="
-          preflight_output="$(timeout 30 "$IMPORTER_PHP" "$IMPORTER_PATH" preflight "${BASE}&directory=${DIR}" /tmp/e2e-smoke --secret="${SECRET}" 2>&1)" || {
+          preflight_output="$(timeout 30 "$IMPORTER_PHP" "$IMPORTER_PATH" preflight "${BASE}&directory=${DIR}" --state-dir="${STATE_DIR}" --fs-root="${FS_ROOT}" --secret="${SECRET}" 2>&1)" || {
             status=$?
             echo "$preflight_output" | head -50
             echo "EXIT: $status"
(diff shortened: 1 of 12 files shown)


.github/workflows/e2e.yml
```diff
@@ -29,14 +29,30 @@ jobs:
...
     needs: build-phar
     runs-on: ubuntu-22.04
     timeout-minutes: 20
     strategy:
       fail-fast: false
       matrix:
-        php: ['7.4', '8.0', '8.1', '8.2', '8.3', '8.4', '8.5']
+        include:
+          - exporter_php: '7.2'
+            importer_php: '8.2'
```
fredrikekelund, line 41: The importer still only supports PHP>=7.4, which is why I separated `exporter_php` and `importer_php` versions

.github/workflows/e2e.yml
```diff
@@ -92,14 +117,29 @@ jobs:
...
 
           echo "=== Importer files-pull ==="
-          timeout 30 php "$IMPORTER_PATH" files-pull "${BASE}&directory=${DIR}" /tmp/e2e-smoke --secret="${SECRET}" 2>&1 | head -50 || echo "EXIT: $?"
+          files_pull_output="$(timeout 30 "$IMPORTER_PHP" "$IMPORTER_PATH" files-pull "${BASE}&directory=${DIR}" --state-dir="${STATE_DIR}" --fs-root="${FS_ROOT}" --secret="${SECRET}" 2>&1)" || {
+            status=$?
+            echo "$files_pull_output" | head -50
+            echo "EXIT: $status"
+            exit "$status"
+          }
+          echo "$files_pull_output" | head -50
```
fredrikekelund, line 139: The old testing command ended with `head -50 || echo "EXIT: $?"`, meaning we would only echo `EXIT: $STATUS` if `head` failed (which it basically never does). Importantly, the script would also not actually exit – it'd just echo `EXIT: 1`.

These changes trigger an actual failure if an error happens here.

phpstan-baseline.neon
```diff
@@ -40,11 +40,6 @@ parameters:
 			count: 1
 			path: packages/reprint-exporter/src/export.php
 
-		-
-			message: "#^Strict comparison using \\!\\=\\= between null and null will always evaluate to false\\.$#"
-			count: 1
-			path: packages/reprint-exporter/src/class-file-tree-producer.php
-
```
fredrikekelund, line 47: `class-file-tree-producer.php` changed in a way that there is no longer an error to suppress

.github/workflows/static-analysis.yml
```diff
@@ -22,3 +22,44 @@ jobs:
...
+          php-version: "8.2"
+          extensions: pdo, pdo_mysql, json, hash, zlib, mbstring
+
+      - run: composer install --no-interaction --prefer-dist
+
+      # Checks exporter compat at PHP 7.2 (shared hosting) and importer at 7.4,
+      # via two PHPCompatibility runs (see lint:php:compat in composer.json).
+      - run: composer lint:php:compat
+
+  syntax-72:
```
adamziel, line 45: Good call! This will prevent drift

PR #357, opened 2026-07-16
[Push] Expose authenticated push operations without buffering uploads

adamziel:
Authenticated push can write a site into the hosting platform's real document root without buffering uploads. `ABSPATH` may point at shared WordPress core, so it is not a safe default for the files the site actually serves.

The exporter adds authenticated create, upload, status, commit, and bounded remove operations. Upload passes `php://input` directly to the multipart processor and reads one bounded fragment at a time. Repeated create validates the existing session before returning success. Each HTTP endpoint now names every public success and failure field instead of forwarding session return arrays or exception context. The receiver preserves configured exclusions and uses `push-create.lock` to coordinate every create and bounded remove step; an unfinished removal tombstone continues to block recreation between steps.

The WordPress entry point accepts the complete trusted options array through the early `site_export_api_options` filter. A hosting platform can set `docroot`, private storage, exclusions, request budgets, and authentication there; a direct `lib.php` embedder passes the same array to `_site_export_handle_api_request()`. `DOCUMENT_ROOT` remains the push default, while `ABSPATH` remains the pull default. The multipart parser also reads through the closing boundary to EOF, rejects trailing data, and reports when that byte instead crosses the decoded request-body limit.

Push hosts must set `display_errors=Off` at PHP startup; PHP or the web server may otherwise emit a non-protocol response before exporter code runs. A bare HTTP 413 is still accepted by the sender and lowers its learned request ceiling. This receiver must not ship without the backend write-authorization gate in #364.

Verified with the focused 116-test push/HTTP suite, PHPStan, PHP 7.2+ compatibility checks for the changed production files, and `git diff --check`.

### DIFF
--- a/packages/reprint-exporter/src/class-push-endpoints.php
+++ b/packages/reprint-exporter/src/class-push-endpoints.php
@@ -161,12 +161,14 @@ final class Site_Export_Push_Endpoints {
      *
      * This operation is idempotent for the same push session ID and immutable
      * server configuration. A successful response reports the independent
-     * multipart-part and decoded request-body limits the sender must apply.
+     * multipart-part and decoded request-body limits the sender must apply,
+     * plus the normalized server-owned excluded paths stored for the session.
      *
      * Emits HTTP 200 with:
      *
      *     {status:"created", push_session_id:string,
-     *      max_part_bytes:int, post_max_bytes:int|null}
+     *      max_part_bytes:int, post_max_bytes:int|null,
+     *      excluded_paths_b64:string[]}
      *
      * @param array $config {
      *     Create request parameters.
@@ -194,6 +196,7 @@ final class Site_Export_Push_Endpoints {
                 'push_session_id' => $push_session_id,
                 'max_part_bytes' => $this->maximum_part_bytes,
                 'post_max_bytes' => $this->post_max_bytes,
+                'excluded_paths_b64' => array_map('base64_encode', $this->excluded_paths),
             ]);
         } catch (Throwable $exception) {
             $this->respond_to_failure($exception);
@@ -537,6 +540,9 @@ final class Site_Export_Push_Endpoints {
      */
     private function respond(int $http_code, array $body): void {
         http_response_code($http_code);
+        header('Cache-Control: no-store, no-cache, must-revalidate, max-age=0');
+        header('Pragma: no-cache');
+        header('Expires: 0');
         header('Content-Type: application/json');
         $json = json_encode($body);
         if ($json === false) {
--- a/packages/reprint-exporter/src/class-push-endpoints.php
+++ b/packages/reprint-exporter/src/class-push-endpoints.php
@@ -1,7 +1,7 @@
 <?php
 
 // phpcs:disable WordPress.Security.EscapeOutput.ExceptionNotEscaped,WordPress.Security.EscapeOutput.OutputNotEscaped -- Authenticated protocol responses are JSON, never HTML output.
-// phpcs:disable WordPress.Security.ValidatedSanitizedInput -- HTTP grammar and receiver validation require the exact request bytes.
+// phpcs:disable WordPress.Security.ValidatedSanitizedInput -- Multipart parsing and request-body byte accounting require the request bytes unchanged.
 
 use function WordPress\Reprint\Exporter\assert_valid_path;
 use function WordPress\Reprint\Exporter\normalize_excluded_paths;
@@ -46,7 +46,7 @@ final class Site_Export_Push_Endpoints {
     private $post_max_bytes;
 
     /**
-     * Configures the server-owned push roots and request limits.
+     * Stores the server-supplied push directories, path policy, and limits.
      *
      * Missing numeric options use bounded defaults. Present numeric options
      * must be positive so a malformed server configuration cannot silently
@@ -56,7 +56,7 @@ final class Site_Export_Push_Endpoints {
      *     Trusted endpoint configuration.
      *
      *     @type string $reprint_directory Required private reprint directory.
-     *     @type string $docroot Required managed document-root directory.
+     *     @type string $docroot Required document-root directory.
      *     @type list<string> $excluded_paths Required document-root-relative
      *         paths which push must preserve.
      *     @type int|float|string $maximum_part_bytes Maximum Content-Length
@@ -162,13 +162,15 @@ final class Site_Export_Push_Endpoints {
      * This operation is idempotent for the same push session ID and immutable
      * server configuration. A successful response reports the independent
      * multipart-part and decoded request-body limits the sender must apply,
-     * plus the normalized server-owned excluded paths stored for the session.
+     * plus the normalized excluded paths stored in push metadata.
      *
-     * Emits HTTP 200 with:
+     * The HTTP 200 response has these fields:
      *
-     *     {status:"created", push_session_id:string,
-     *      max_part_bytes:int, post_max_bytes:int|null,
-     *      excluded_paths_b64:string[]}
+     * - `status`: `created`.
+     * - `push_session_id`: the caller-supplied ID.
+     * - `max_part_bytes`: the maximum multipart-part Content-Length.
+     * - `post_max_bytes`: the decoded request-body limit, or null.
+     * - `excluded_paths_b64`: the stored excluded paths in base64.
      *
      * @param array $config {
      *     Create request parameters.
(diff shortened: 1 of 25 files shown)


packages/reprint-exporter/src/class-push-endpoints.php
```diff
@@ -0,0 +1,549 @@
...
+            $push_session = Site_Export_Push_Session::create(
+                $this->reprint_directory,
+                $this->docroot,
+                $this->excluded_paths,
+                $push_session_id
+            );
+            // create() deliberately defers validation when the directory
+            // already exists. An endpoint success must not bless stale or
+            // incompatible durable state, so validate it under the push lock.
+            $push_session->get_status();
```
adamziel: This sounds opaque, we need to explain it more clearly. What's the specific data flow? Problem? And why does create() defer it? Is that a useful behavior?

packages/reprint-exporter/src/class-push-endpoints.php
```diff
@@ -0,0 +1,555 @@
...
+     *
+     * Emits HTTP 200 with one of these `last_change` branches, or null for an
+     * empty request:
+     *
+     *     {status:"accepted", push_session_id:string, changes_accepted:int,
+     *      last_change:
+     *        {path_b64:string,state:"partial"|"complete",type:"file",accepted_bytes:int}
+     *        | {path_b64:string,state:"complete",type:"directory"|"symlink",accepted_bytes:0}
+     *        | {state:"partial"|"complete",type:"delete-list",accepted_bytes:int}
+     *        | null}
```
adamziel: this needs to be better formatted to be readable

packages/reprint-exporter/src/class-push-endpoints.php
```diff
@@ -0,0 +1,555 @@
...
+                $this->reprint_directory,
+                $this->docroot,
+                $push_session_id,
+                $this->excluded_paths
+            );
+            $commit = $push_session->commit($this->maximum_commit_entries);
+            $this->respond(200, array_merge([
+                'status' => 'accepted',
+                'push_session_id' => $push_session_id,
+            ], $commit));
```
adamziel: hm, it's somewhat unclear what's the shape returned by commit() and I don't like that we automatically expose it from the API. what if it changes? Let's make the API surface explicit

packages/reprint-exporter/src/class-push-endpoints.php
```diff
@@ -0,0 +1,555 @@
...
+            $removed = Site_Export_Push_Session::remove(
+                $this->reprint_directory,
+                $this->docroot,
+                $push_session_id,
+                $this->excluded_paths
+            );
+            $this->respond(200, [
+                'status' => 'accepted',
+                'push_session_id' => $push_session_id,
+                'removed' => $removed,
```
adamziel: Ditto, Let's make the API surface explicit

packages/reprint-exporter/src/class-push-session.php
```diff
@@ -2232,29 +2292,60 @@ private function work_deletes_are_complete(): bool {
...
             }
         }
     }
 
+    /**
+     * Validates path grammar without applying the excluded-path policy.
+     *
+     * @param string $path Document-root-relative raw path bytes.
+     */
+    private function validate_path_syntax(string $path): void {
```
adamziel: This is a terrible name. Path syntax is "any bytes except for \0" and thi doesn't validate that at all. We need a better name, such as `assert_path_not_reserved()` or so

packages/reprint-exporter/src/class-push-session.php
```diff
@@ -2232,29 +2292,60 @@ private function work_deletes_are_complete(): bool {
...
+     * Validates one implicit structural work-tree directory.
+     *
+     * A structural directory is traversed only to reach requested descendant
+     * work. It may therefore be an ancestor of an excluded path when the work
+     * lies in an unrelated sibling, but it must never equal or descend from an
+     * excluded path itself.
+     *
+     * @param string $path Document-root-relative structural directory path.
+     */
+    private function validate_structural_path(string $path): void {
```
adamziel: I don't want the word "structural" to be a part of this repository. Let's find a specific word that communicates what this is about.

packages/reprint-exporter/src/class-push-session.php
```diff
@@ -2477,6 +2568,50 @@ private function file_size(string $path): int {
...
+     * Creates or validates the directory shared by every push session.
+     *
+     * Create and remove both establish this directory before acquiring the
+     * lifecycle lock. This lets an idempotent remove coordinate with a create
+     * even when no push session or tombstone currently exists.
+     *
+     * @param string $reprint_directory Canonical private reprint directory.
+     * @return string Canonical push parent directory.
+     */
+    private static function require_push_parent_directory(string $reprint_directory): string {
```
adamziel: I don't like the word `require` here. Let's find a way to express this intention more specifically.

reprint-exporter-wp/lib.php
```diff
@@ -277,8 +359,130 @@ function _site_export_handle_api_request(array $options = []): void {
...
+                );
+            }
+            $canonical_docroot = realpath($configured_docroot);
+            if ($canonical_docroot === false || !is_dir($canonical_docroot)) {
+                throw new Site_Export_Push_Configuration_Exception(
+                    'Push endpoints require docroot or DOCUMENT_ROOT to name an existing directory; observed '
+                    . json_encode($configured_docroot) . '.'
+                );
+            }
+            $docroot = $canonical_docroot === '/' ? '/' : rtrim($canonical_docroot, '/\\');
```
adamziel, line 388: the platform must be able to override this.

packages/reprint-exporter/src/class-push-endpoints.php
adamziel: Agreed. `Site_Export_Push_Session::create()` now validates a reopened directory’s layout, immutable configuration, and same-filesystem relationship under the push lock while the create/remove lock is still held. The endpoint only calls `create()`, so a successful create owns that complete postcondition instead of adding a second validation step.

packages/reprint-exporter/src/class-push-endpoints.php
adamziel: Expanded this union into a field-by-field list. I applied the same readable response format to all five endpoint docs and the push protocol documentation.

packages/reprint-exporter/src/class-push-endpoints.php
adamziel: The commit endpoint now maps only `phase`, `send_next_request`, and `entries_processed` into its response, and the endpoint test asserts that exact object. I generalized this boundary to upload, status, and failure responses so lower-level return arrays and exception context cannot silently expand the HTTP API.

packages/reprint-exporter/src/class-push-endpoints.php
adamziel: The remove endpoint now names the result `$remove_complete` and constructs only `status`, `push_session_id`, and `removed`. Exact endpoint tests cover the initial false result, bounded retries, repeated removal, and the final true result.

packages/reprint-exporter/src/class-push-session.php
adamziel: Renamed this to `assert_path_not_reserved()`. I also split the two excluded-path policies into `assert_path_is_not_excluded()` for work ancestor traversal and `assert_path_does_not_overlap_excluded_paths()` for requested values, so each name states the condition it enforces.

packages/reprint-exporter/src/class-push-session.php
adamziel: Removed “structural” from the push source, docs, and tests. The code now distinguishes document-root ancestor directories from work ancestor directories, which names the actual role of each path.

packages/reprint-exporter/src/class-push-session.php
adamziel: Renamed this to `create_push_sessions_directory()` because it creates the shared directory and returns its canonical path. I generalized the naming cleanup to the new endpoint helpers too: they now assert the request method and read the push session ID rather than vaguely “requiring” them.

reprint-exporter-wp/lib.php
adamziel: The bundled WordPress `index.php` now applies an early `site_export_api_options` filter and forwards the complete trusted options array. A real-router test enters through `index.php` and proves that the platform’s `docroot` overrides `DOCUMENT_ROOT`; direct `lib.php` embedders pass the same array explicitly.

PR #372, opened 2026-07-20
Refactor: Extract importer multipart stream parser module

atanas-dev:
Start moving pieces outside of import.php to slowly chip away at its size.

### DIFF
--- a/packages/reprint-importer/src/import.php
+++ b/packages/reprint-importer/src/import.php
@@ -38,6 +38,12 @@ require_once __DIR__ . '/lib/mysql-query-stream/load.php';
 // Load WordPress function stubs (needed by wp-php-toolkit outside WordPress)
 require_once __DIR__ . '/lib/wp-stubs.php';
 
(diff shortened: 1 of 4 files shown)


packages/reprint-importer/src/import.php
```diff
@@ -38,6 +38,12 @@
 // Load WordPress function stubs (needed by wp-php-toolkit outside WordPress)
 require_once __DIR__ . '/lib/wp-stubs.php';
 
+// Streaming protocol parsers.
+require_once __DIR__ . '/lib/protocol/class-multipart-stream-parser.php';
+if (!class_exists('MultipartStreamParser', false)) {
+    class_alias(\Reprint\Importer\Protocol\MultipartStreamParser::class, 'MultipartStreamParser');
+}
```
adamziel, line 45: We could move this to autoload. Not a blocker here, just an idea for later. We'll also need to rename these from "importer" and "exporter" as push support changes these roles. Good work here!

PR #374, opened 2026-07-20
Refactor: Extract importer file placement rules module

atanas-dev:
Start moving pieces outside of import.php to slowly chip away at its size.

### DIFF
--- a/packages/reprint-importer/src/lib/file-sync/class-file-placement-rules.php
+++ b/packages/reprint-importer/src/lib/file-sync/class-file-placement-rules.php
@@ -1,0 +1,669 @@
+<?php
+
+namespace Reprint\Importer\FileSync;
+
+use InvalidArgumentException;
+use RuntimeException;
+use function WordPress\Filesystem\wp_join_unix_paths;
+use function WordPress\Reprint\Exporter\assert_valid_path;
+use function WordPress\Reprint\Exporter\normalize_path;
+use function WordPress\Reprint\Exporter\path_is_within_root;
+
+/**
+ * Pure file-placement rules for files-pull.
+ *
+ * This class decides where remote source paths land locally and which source
+ * paths should be traversed. It deliberately accepts plain arrays and scalar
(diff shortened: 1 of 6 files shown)


packages/reprint-importer/src/lib/file-sync/class-file-placement-rules.php
```diff
@@ -0,0 +1,669 @@
...
+
+/**
+ * Pure file-placement rules for files-pull.
+ *
+ * This class decides where remote source paths land locally and which source
+ * paths should be traversed. It deliberately accepts plain arrays and scalar
+ * values so CLI parsing, persisted state, index files, and audit output remain
+ * outside the business rules.
+ */
+final class FilePlacementRules
```
adamziel, line 20: I'm not convinced about this additional abstraction. I get the desire to shrink the importer class. At the same time, I'd hesitate to change `FilePlacementRules` in the future without tracing the importer very carefully, and that tells me that this particular split increases the total complexity and cognitive load of the codebase. That, in turn, tells me we may be better off keeping this particular code in the importer class.

The volume of code certainly requires some getting used to. I have been adamantly against such big classes for years, until @dmsnell and I built the [HTML parser for WordPress core](https://raw.githubusercontent.com/WordPress/wordpress-develop/4ef910b1c3a6f034a681b3c1bd1c05f65e30905a/src/wp-includes/html-api/class-wp-html-tag-processor.php#L411-L5077). Most of the work there is done by two very bulky parent/child classes that are approximately 4500 and 6000 lines long. If we were to do it again, I'd be very tempted to go for a single class with 10k lines of code because to reduce the cognitive load of `parent::` calls and overrides. We've extracted some parts of that into smaller classes, and I don't think there are any parts left that, if separated, would make it easier to work with.

I'm happy with extracting independent responsibilities from the importer class. I just wouldn't extract the integral logic. Complexity and scattered interdependent logic is a bigger challenge than volume.

adamziel (changes requested):

atanas-dev:
Closing in favor of a different approach.

PR #378, opened 2026-07-20
[Push] Serve every push request without loading WordPress

adamziel:
All `files-push` requests now use a PHP route that does not boot WordPress, so a new push can replace files which broke WordPress startup.

## Background

A push can install a plugin which throws while WordPress starts. If `push_create` goes through WordPress, the next push cannot start and replace that plugin.

## This change

Create, upload, status, commit, and remove now use one push URL. The WordPress route serves pull requests only and rejects every `push_*` request.

`push_create` uses the connection token. It returns a random secret for that push session. The other four operations use the session secret at the same URL.

There is still one implementation of each push operation. `Site_Export_HTTP_Server::serve_push()` authenticates the request and dispatches it to the existing `Site_Export_Push_Endpoints` methods. The old WordPress-side push setup and dispatch are removed.

The bundled URL ends in:

```text
/wp-content/plugins/reprint-exporter/push.php
```

The Reprint Exporter settings screen shows the exact URL. Pass it to `files-push` instead of `?reprint-api`:

```bash
php reprint.phar files-push "https://example.com/wp-content/plugins/reprint-exporter/push.php" \
  --state-dir="/path/to/reprint-state" \
  --fs-root="/path/to/local-tree" \
  --secret="the-target-connection-token"
```

The bundled route reads `DOCUMENT_ROOT`, uses its private sibling reprint directory, and excludes its own plugin directory from push.

Enabling push writes the connection token to the mode-0600 `<reprint-directory>/.reprint/push-config.php`. The file returns a PHP array and prints nothing. Each push session keeps its secret and policy in a mode-0600 `push-metadata.php`, which also returns an array without printing it. Removing `push-config.php` blocks new sessions without breaking a session already in progress.

Some hosts block direct PHP below `wp-content/plugins`, or use different server-owned paths. Those hosts must expose another PHP URL which does not load WordPress:

```php
require_once '/srv/reprint-exporter/vendor/autoload.php';

$connection_secret = ($_GET['endpoint'] ?? null) === 'push_create'
    ? Site_Export_HTTP_Server::load_push_connection_secret('/srv/reprint/push-config.php')
    : null;

Site_Export_HTTP_Server::serve_push([
    'connection_secret' => $connection_secret,
    'docroot' => '/srv/www/public',
    'reprint_directory' => '/srv/reprint',
    'excluded_paths' => ['reprint-push.php'],
]);
```

Give that URL directly to `files-push`. The WordPress `site_export_api_options` filter no longer configures push.

## Testing

Push a plugin which makes the WordPress route return HTTP 500. Replace the local plugin with a safe version and run `files-push` again with the same state directory. Confirm that the second run creates a new push session and repairs the site through `push.php`.

### DIFF
--- a/markdown/PUSH-SYNC.md
+++ b/markdown/PUSH-SYNC.md
@@ -54,25 +54,29 @@ size. The secret travels in no URL and no body, so it never lands in an
 access log.
 
 Authentication does not grant write authority. Connection tokens are
-download-only by default, including tokens that predate push endpoints. Except
-for the bounded recovery case below, every `push_*` operation requires current
-push authorization before upload data is read, a push directory is created, or
-the document root changes; custom authentication uses the same gate.
+download-only by default, including tokens that predate push endpoints.
+`push_create` requires current push authorization, but it does not boot
+WordPress. The bundled push route reads the connection secret from a private
+PHP configuration file and authenticates the signed request itself.
 
-Revocation does not abandon durable recovery state. An authenticated caller
-may keep calling `push_commit` only when that push session already has a durable
-commit checkpoint. Those requests converge the already-started document-root
-mutation; they cannot upload more work, inspect or remove private work, or
-start commit for another push session. Other push operations return HTTP 403
-with `reason: "push_disabled"`. This keeps token rotation or a managed policy
-change from stranding a partial commit and its maintenance marker.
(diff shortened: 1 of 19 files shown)


adamziel:
Superseded by #460, which keeps the existing connection-token and authorization model while adding the standalone route on current trunk.

PR #381, opened 2026-07-21
Refactor: Extract importer file index module

atanas-dev:
TBD

### DIFF
--- a/packages/reprint-importer/src/import.php
+++ b/packages/reprint-importer/src/import.php
@@ -12,6 +12,10 @@
 
 use function WordPress\Filesystem\wp_join_unix_paths;
 use Reprint\Importer\FileSync\FilePlacementRules;
+use Reprint\Importer\Index\IndexFileSorter;
+use Reprint\Importer\Index\IndexLineParser;
+use Reprint\Importer\Index\IndexPathPrefixMatcher;
+use Reprint\Importer\Index\IndexStore;
 use Reprint\Importer\Tuning\AdaptiveTuner;
 use function WordPress\Reprint\Exporter\assert_valid_path;
 use function WordPress\Reprint\Exporter\normalize_path;
@@ -66,6 +70,9 @@ require_once __DIR__ . '/lib/target-runtime/load.php';
 
 // External merge sort for large index files when exec() is unavailable
 require_once __DIR__ . '/lib/external-merge-sort.php';
+
+// Sorted JSONL file indexes for local and remote file manifests.
+require_once __DIR__ . '/lib/index/load.php';
 
 // Terminal progress rendering (spinner, progress lines, lifecycle messages)
 require_once __DIR__ . '/lib/terminal-progress/class-terminal-progress.php';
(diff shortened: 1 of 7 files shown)


packages/reprint-importer/src/lib/index/class-index-line-parser.php
```diff
@@ -0,0 +1,70 @@
+<?php
+
+namespace Reprint\Importer\Index;
+
+use RuntimeException;
+use function WordPress\Reprint\Exporter\assert_valid_path;
+
+final class IndexLineParser
```
adamziel, line 8: This class seems to exist solely as a namespace for the two methods. It doesn't use any internal state and it cannot be extended. Should we just go with two standalone functions? I'm worried about creating a proliferation of small classes.

packages/reprint-importer/src/import.php
```diff
@@ -597,19 +584,37 @@ public function __construct(
...
+
+    private function index_store(): IndexStore
+    {
+        if ($this->index_store === null) {
+            $this->index_store = new IndexStore(
+                $this->index_file,
+                $this->index_updates_file,
+                [$this, "audit_log"],
+            );
         }
```
adamziel, line 600: Before this PR, I could read this 12 lines long method and understand everything about it. With this PR, I need to read multiple methods, another class, and mentally keep track of additional mutable state. I'm not convinced this is helpful

packages/reprint-importer/src/import.php
```diff
@@ -6759,380 +6752,35 @@ private function remove_local_path_without_following_symlinks(
...
-            return null;
-        }
-        while (($line = fgets($handle)) !== false) {
-            $parsed = $this->parse_index_line($line);
-            if ($parsed !== null) {
-                return $parsed;
-            }
-        }
-        return null;
+        return $this->index_store()->readIndexLine($handle);
```
adamziel, line 7053: Pre-existing methods in this repo already use snake_case. Methods introduced in the new classes use camelCase. Let's stick to snake_case since that's what WordPress does.

packages/reprint-importer/src/import.php
```diff
@@ -597,19 +584,37 @@ public function __construct(
...
+            );
         }
-        $count = 0;
-        while (fgets($handle) !== false) {
-            $count++;
+
+        return $this->index_store;
+    }
+
+    private function remote_index_prefix_matcher(): IndexPathPrefixMatcher
```
adamziel, line 605: When we're introducing new code, let's make sure it comes with informative doc blocks that explain in simple terms to a new reader what the code does. It makes navigating the codebase much easier without analyzing all the calls. Unfortunately, I haven't found a good way to get LLMs to consistently do that—they often write trivialities or rewrite the code in plain english. The best prompt I found so far is "document code like Linus Torvalds," but I still often spend time rewriting and expanding these comments to explain the intent, provide code examples etc.

One of the best documented classes I know is [WP_HTML_Processor](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/html-api/class-wp-html-processor.php), largely thanks to @dmsnell. I always keep that in mind as an ideal model.

packages/reprint-importer/src/import.php
```diff
@@ -6759,380 +6752,35 @@ private function remove_local_path_without_following_symlinks(
...
-            throw new RuntimeException("Invalid index path (base64 decode failed)");
-        }
-        assert_valid_path($path, "index path");
-        return [
-            "path" => $path,
-            "ctime" => (int) ($data["ctime"] ?? 0),
-            "size" => (int) ($data["size"] ?? 0),
-            "type" => (string) ($data["type"] ?? "file"),
-        ];
+        return IndexLineParser::parse($line);
```
adamziel, line 7027: We could also remove the entire private function and just call a global `parse_index_line( $line )` directly. That being said, it's a pretty small function. Are we gaining much by removing it one step from the class that also determines the shape of that index? Looking at this a few lines below, I don't think OOP is serving us well here:

> $this->index_store()->beginUpdates();
> $this->index_updates_file = $this->index_store()->updatesFile();

Perhaps we could extract the entire indexing stage of the pipeline as the sole new class with just one responsibility?

adamziel (changes requested):

packages/reprint-importer/src/lib/index/class-index-file-sorter.php
```diff
@@ -0,0 +1,190 @@
...
+        $mem_limit = ($mem_limit_raw === "-1" || $mem_limit_raw === "" || $mem_limit_raw === "0")
+            ? 0
+            : parse_size($mem_limit_raw);
+        $mem_used = memory_get_usage(true);
+        $available = $mem_limit > 0
+            ? (int) (($mem_limit - $mem_used) * 0.6)
+            : 256 * 1024 * 1024;
+
+        $sorter = new \ExternalMergeSort(
+            new IndexLinePathKeyExtractor(),
```
adamziel, line 51: Ditto as other comments – we're swapping a tiny inline function for another class living in another file that delegates work into yet another class, I'd say this is adding cognitive effort, not decreasing it.

atanas-dev:
Closing in favor of a different approach.

PR #384, opened 2026-07-21
Guard the vendor autoload include against symlink double-loading

epeicher:
## What it does

Requires the plugin's Composer autoloader by its `realpath()`, so the same physical `autoload.php` can no longer be included twice through two different path strings.

## Rationale

When `wp-content` is a symlink (the WP Cloud layout), the plugin's files are reachable through two path families — the symlink path and the realpath. With `opcache.revalidate_path=0` (the default), `require_once` does not dedupe across the two strings, so a cross-family double include of `autoload.php` re-declares the `ComposerAutoloaderInit*` class:

```
PHP Fatal error: Cannot redeclare class ComposerAutoloaderInitc5be971...
  previously declared in /srv/<site>/wp-content/plugins/site-export/vendor/composer/autoload_real.php   ← symlink family
  in                     /tmp/<content>/wp-content/plugins/site-export/vendor/composer/autoload_real.php ← realpath family
```

The export endpoint answers HTTP 500 and, in CI, `import-38-flatten-wpcloud` cascades (files-sync fails, nothing lands in the fs root, flat-docroot reports "ABSPATH not found"). Which leg fails varies with per-file opcache cache-key state — it hit PHP 8.1 twice, then 8.4 — the same class of symlink double-load afc8318 (#148) fixed inside `export.php`.

A temporarily instrumented CI run (since reverted) captured the fatal in flight and pinned the mechanism:

- the plugin bootstraps once through the symlink family, but **Composer's baked `__DIR__` paths resolve symlinks**, so everything the ClassLoader itself pulls in — classmap classes, and eventually the colliding second `autoload.php` include — runs through the **realpath family**;
- a symbol guard (`function_exists` on a Composer `files` entry, the first attempt in this PR) is unsound: the trace showed the loader re-entered mid-`getLoader()`, when the `ComposerAutoloaderInit*` class is already declared but the `files` entries have not run yet.

Requiring the realpath puts this loader's include in the same family Composer already uses, so `require_once` dedupes natively whichever side runs first — no reliance on load-order symbols, no dependence on who the second includer is.

## Testing instructions

The failure is opcache- and version-dependent, so the E2E matrix is the practical test: `import-38-flatten-wpcloud` exercises exactly this layout. With the fix, the full matrix went green ([run](https://github.com/WordPress/reprint/actions/runs/29852123974) failed on 8.4 before the fix; the follow-up runs pass) — the flake had fired on every recent PR run, including twice on the same leg of #383. `php -l reprint-exporter-wp/lib.php` passes; behaviour is unchanged when no symlinks are involved (`realpath` is then the same path).

### DIFF
--- a/reprint-exporter-wp/index.php
+++ b/reprint-exporter-wp/index.php
@@ -9,6 +9,50 @@
  * License URI: https://www.gnu.org/licenses/gpl-2.0.html
  */
 
+// ── TEMPORARY E2E DIAGNOSTIC — REVERT BEFORE MERGE ──────────────────────────
+// Traces the double-bootstrap behind the intermittent wpcloud-flatten fatal
+// ("Cannot redeclare class ComposerAutoloaderInit*"). Gated to the wpcloud
+// E2E vhost so the rest of the suite stays quiet.
+if (
+    getenv('SITE_EXPORT_TEST_MODE') &&
+    strpos($_SERVER['DOCUMENT_ROOT'] ?? '', 'wpcloud') !== false
+) {
+    $stu_diag_trace = array_map(
+        static function ($f) {
+            return ($f['file'] ?? '?') . ':' . ($f['line'] ?? 0);
+        },
+        debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS)
+    );
+    error_log(
+        '[E2E-DIAG] index.php __FILE__=' . __FILE__
+        . ' uri=' . ($_SERVER['REQUEST_URI'] ?? '-')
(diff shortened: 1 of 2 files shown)


epeicher:
Closing this PR: CI traces showed the fatal is not fixable at the plugin's include site. Both attempts (symbol guard, then requiring the autoloader by realpath) still reproduced the crash, because with `opcache.revalidate_path=0` the same file reached through a symlink and its realpath gets two opcache identities, so `require_once` can't dedupe them. The real fix belongs in the E2E infrastructure (e.g. `opcache.revalidate_path=1` for the FPM pools), not in `lib.php`.

PR #386, opened 2026-07-21
[Push] Advance one local baseline through pull and push batches

adamziel:
Compatible file-only pulls and committed `files-push` operations now advance the same `previous_local_index.jsonl` only for paths they applied. `files-diff` stops reporting transferred work while local additions, edits, and deletions elsewhere remain pending.

## Background

`files-diff` compares the current local tree with the pair's previous local index. Before this PR, `files-push` replaced that baseline with its complete plan-start local index after target commit, while pull mutations advanced only the import index. Pull and push therefore did not maintain one durable baseline, and a resumed sender could overwrite paths advanced by a compatible pull between sender processes.

## This change

A compatible full file-only pull seeds a missing previous local index before its first local mutation. Each durable WAL batch then publishes the applied paths into both the import index and the previous local index before clearing the batch. A compatible partial pull advances an existing baseline but cannot seed one; an incompatible pull removes the baseline before mutating the tree.

After target commit, `files-push` merges only its committed push and delete operations, plus their required ancestors, into the latest baseline. It no longer replaces that baseline with the plan-start snapshot.

One state-directory `.reprint.lock` excludes other local Reprint commands throughout the lifecycle. An empty `.import-index-updates.wal` remains as the `files-pull` lifecycle marker, and a non-empty WAL is replayed after interruption. Pull and push use the same streaming index merger, so large indexes remain single-pass and bounded in memory.

## Testing

Run a compatible full file-only pull, make a local change, then run a compatible partial pull for another path. Confirm that `files-diff` still reports the local change but not the pulled path.

Complete a `files-push` and confirm that its committed paths leave `files-diff` while unrelated pending paths remain. Interrupt a pull with a WAL batch open, resume it, and confirm the completed batch is applied once to both indexes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -679,7 +679,7 @@ truncated or rotated, so it provides a complete history of the migration.
 ```
 [2025-01-15 10:30:01] VOLATILE | path=/srv/htdocs/wp-content/debug.log | count=1
 [2025-01-15 10:30:05] VOLATILE CLEARED | path=/srv/htdocs/wp-content/debug.log
-[2025-01-15 10:31:12] FILE DELETE | .import-index-updates.jsonl
+[2025-01-15 10:31:12] FILE TRUNCATE | .import-index-updates.wal | WAL batch applied
 ```
 
 Pass `--verbose` to also print audit log entries to the console as they happen.
--- a/markdown/PUSH-SYNC.md
+++ b/markdown/PUSH-SYNC.md
@@ -7,9 +7,9 @@ the end maps it to a PR stack.
 
 ## Shape of a push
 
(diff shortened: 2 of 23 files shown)


adamziel:
Replaced by this three-PR stack:

1. #404 — one state-directory-wide Reprint process lock
2. #405 — files-pull index-update WAL
3. #406 — shared previous local index advanced by compatible pulls and committed pushes

The third branch contains the resolved final tree from this PR, with the lock and WAL isolated below it. All three replacement PRs are mergeable; CI is running on the stack.

PR #395, opened 2026-07-27
[Push] Add pull conflict policies for local and remote changes

adamziel:
File-only delta pulls now stop before replacing a path changed locally and remotely, with explicit `--on-conflict=remote-wins` and `--on-conflict=our-wins` alternatives.

## This change

A compatible `files-pull` or `pull-files` compares the current remote index with `.import-index.jsonl` and compares a fresh local index with `previous_local_index.jsonl`. The two change streams are sorted and intersected without loading either tree into memory. A remote deletion or file replacement also conflicts with local changes inside that path.

`stop` is the default and completes the comparison before the diff changes the local tree. `remote-wins` applies the remote change. `our-wins` leaves the local path untouched and does not update either index for that remote change, so the remote change remains pending for a later pull and the local change remains pending for `files-diff` and `files-push`.

Stacked on #386.

## Testing

Complete a file-only pull, change a local 10-byte file to 11 bytes, and change the same file on the source. Confirm the default stops with the local file still at 11 bytes. Resume with `our-wins` and confirm `files-diff` reports the local edit. Repeat with `remote-wins` and confirm the source content replaces the local edit.


packages/reprint-importer/src/import.php
```diff
@@ -408,6 +384,9 @@ class ImportClient
      */
     private $include_caches = false;
```
adamziel, line 385: This would be better as `pull_cached_files` or so. .git and node_modules are not quite caches, though, so we might need a better name still.

packages/reprint-importer/src/import.php
```diff
@@ -3381,8 +3338,16 @@ private function run_files_sync_pipeline(): void
                 }
             }
             $this->sort_index_file($this->remote_index_file);
-            $this->import_state()->active_resumable_command->current_stage = "diff";
+            $stage =
```
adamziel, line 3341: Is this implicit? Shouldn't it be explicit?

packages/reprint-importer/src/import.php
```diff
@@ -3396,7 +3361,29 @@ private function run_files_sync_pipeline(): void
                 );
             }
             $this->save_state($this->state);
+        }
+
+        if ($stage === "conflicts") {
+            $conflict_count = $this->prepare_files_pull_conflicts();
+            if ($conflict_count === null) {
+                $this->import_state()->active_resumable_command->completion_state = "partial";
+                $this->save_state($this->state);
```
adamziel, line 3370: This reads weird, we're modifying one state and saving another. Is there an implicit dependency between them?

packages/reprint-importer/src/import.php
```diff
@@ -6861,6 +6848,62 @@ private function diff_indexes_and_build_fetch_list(): bool
         $diff = $this->import_state()->diff;
         $remote_offset = $diff->remote_offset;
         $local_after = $diff->local_after;
+        $conflict_offset = $diff->conflict_offset;
+        $conflicts_handle = null;
+        $conflict_path = null;
+        $next_conflict_offset = $conflict_offset;
+        if (
```
adamziel, line 6855: I could really use an inline comment here explaining what this operation is

packages/reprint-importer/src/import.php
```diff
@@ -6861,6 +6848,62 @@ private function diff_indexes_and_build_fetch_list(): bool
...
+            }
+            fseek($conflicts_handle, $conflict_offset);
+            $conflict_path = stream_get_line(
+                $conflicts_handle,
+                1048576,
+                "\0"
+            );
+            $next_conflict_offset = ftell($conflicts_handle);
+        }
+        $path_has_conflict = function (string $path) use (
```
adamziel, line 6874: I wouldn't mind returning an object instead of setting values by reference.

adamziel: Also I don't love an inline function.

packages/reprint-importer/src/import.php
```diff
@@ -6861,6 +6848,62 @@ private function diff_indexes_and_build_fetch_list(): bool
...
+            $conflicts_handle,
+            &$conflict_path,
+            &$conflict_offset,
+            &$next_conflict_offset
+        ): bool {
+            if (!is_resource($conflicts_handle)) {
```
adamziel, line 6880: Why not trust the input? Are we going to pass invalid values here or what?

PR #408, opened 2026-07-28
Remove NO_ENGINE_SUBSTITUTION SQL header

WPprodigy:
A target host could be using `enforce_storage_engine` to enforce say InnoDB, while a source's DB export could contain some MyISAM tables. With `NO_ENGINE_SUBSTITUTION`, the import fails (unnecessarily).

It seems better overall to allow the import to go through. If there's a true incompatibility due to the engine change (something quite rare), the table schema will be seen as invalid and it'll error out still.

### DIFF
--- a/packages/reprint-exporter/src/class-mysql-dump-producer.php
+++ b/packages/reprint-exporter/src/class-mysql-dump-producer.php
@@ -31,7 +31,7 @@ use PDOStatement;
  * of UPDATE ... SET col = CONCAT(col, chunk) statements.
  *
  * Known limitations:
- * 
+ *
  * - Rows too large to be SELECTed. If a row is larger than max_allowed_packet or the
  *   PHP memory_limit, it won't be exported. The underlying assumption is that WordPress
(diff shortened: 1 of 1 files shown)


adamziel:
@WPprodigy thank you! Let's get this in. MyISAM -> InnoDB sounds absolutely fine. If we want to be extra careful, we should default to `NO_ENGINE_SUBSTITUTION` to prevent InnoDB -> MyISAM which could be disastrous.

WPprodigy:
A `InnoDB -> MyISAM` downgrade would only happen if the target doesn't support InnoDB. Which on  MySQL 8 is impossible. Could be done w/ MariaDB, but that shouldn't really be a thing in practice I imagine.

Preflighting the target environments could be interesting (has come up in some other threads of work too). In this case, it could come in handy by preflighting the target and aborting if InnoDB is not available while the source has InnoDB tables.

I'm not really sure what other mechanism we'd want here. A best-effort import in the target seems best for now. If the target/host disables InnoDB, then it's kind of asking for such trouble 🙃

PR #409, opened 2026-07-29
Move symbols to their own files

atanas-dev:


adamziel:
Could we rebase this on trunk? It seems to be only moving the existing code around and, as such, is a risk-free change that we could easily get in. #394, on the other hand, I'm not convinced about and I'd like to challenge some assumptions in there, and we may end up discussing it for a while. There's no need to block this PR on that.

atanas-dev:
@adamziel done :)

adamziel:
Nice, thank you! I'm merging some wording changes that are creating conflicts here, let's connect synchronously tomorrow to get this merged

PR #431, opened 2026-07-30
Restore Reprint state filenames consumed by WP Cloud

adamziel:
WP Cloud migrations continue to read Reprint's established state files under `--state-dir`.

The deployed `wpcloud-reprint-adapter` reads `.import-state.json` after preflight and during full sync to select the database charset, build URL rewrites, and set `$table_prefix`. #424 moved that file, so those reads silently returned no state. Pull state also stores positions into the index, WAL, and path-list files, so this restores the complete on-disk filename set rather than making one persisted lifecycle span the old and new layouts. The state-directory-wide lock returns to `.reprint.lock` so every Reprint version using this layout serializes on the same file.

The internal lifecycle names from #424 and runtime constant names from #429 remain unchanged.

## Testing

`cd tests && ../vendor/bin/phpunit --colors=never Import`

`cd tests && ../vendor/bin/phpunit --colors=never PushEndpointsTest.php`

`vendor/bin/phpstan analyze --memory-limit=1G`

`vendor/bin/phpcs --standard=PHPCompatibility --runtime-set testVersion 7.4- -ps packages/reprint-importer/src`

`php tests/run.php` in the WP Cloud adapter checkout

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -166,7 +166,7 @@ Entries in `paths_to_remove` under `wp-content/plugins/` also trigger automatic 
 
 ### SQL Streaming Crash Recovery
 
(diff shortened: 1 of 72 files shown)


adamziel:
Closing this approach. The compatibility adjustment belongs in the WP Cloud adapter rather than restoring Reprint's former state filenames.

PR #500, opened 2026-08-06
[Push] Select files-push progress output

adamziel:
`files-push` can now select terminal or JSONL progress explicitly while keeping TTY detection as the default.

This is stacked on #484 and extends the `--progress=auto|tty|jsonl` selector from #492 to `files-push`.

```sh
# Keep the single progress bar when stdout is captured.
reprint files-push "$URL" --state-dir="$STATE_DIR" --fs-root="$FS_ROOT" \
  --secret="$SECRET" --progress=tty

# Keep the existing JSONL stream when running in a terminal.
reprint files-push "$URL" --state-dir="$STATE_DIR" --fs-root="$FS_ROOT" \
  --secret="$SECRET" --progress=jsonl
```

`--progress=auto` preserves the current behavior: terminal output on a TTY and JSONL otherwise. `tty` forces the bar from #484; `jsonl` forces the existing `push_progress` records and final result. Both presentations consume the same sender progress snapshot. The selection belongs to one invocation and is not written to sender state.

This uses a progress selector instead of a general `--format` option because it changes only the live progress presentation, not Reprint's result schema. The shape follows [BuildKit's `--progress=auto|tty|rawjson` selector](https://docs.docker.com/reference/cli/docker/buildx/build/#set-type-of-progress-output---progress), while `jsonl` names the line-delimited format Reprint already emits.

Explicit `tty` and `jsonl` modes reject `--verbose`, so detailed log lines cannot corrupt the selected presentation.

## Testing

Run `files-push --progress=tty` with stdout captured and check that one bar redraws through `Indexing`, `Pushing`, and `Committing`. Run `--progress=jsonl` directly in a terminal and check that each output line is a JSON record. Try either explicit mode with `--verbose` and check that the command rejects the combination before starting a sender.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -174,6 +174,8 @@ During the file fetch phase, progress and heartbeat records include `files_done`
 
 During files-push, progress records include `files_done` and `files_total` together after planning. The completed count advances only at target-confirmed request boundaries and survives resume.
 
+Interactive non-verbose files-push output uses one stage-weighted progress bar for the complete lifecycle. The percentage precedes a label which changes only at major lifecycle stages; while pushing local paths, target-confirmed file bytes appear beside the planned file byte total. These terminal-only details are not added to JSONL or `progress.json`.
+
 ## File Organization
 
 - packages/reprint-server/: Packagist server package (previously reprint-exporter)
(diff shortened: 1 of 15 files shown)


adamziel:
Folded into #484.

PR #537, opened 2026-08-08
[PHP] Share sorted file-index comparison between push and pull

adamziel:
`PushPlan` now delegates its two-index traversal to `FileIndexDiffProcessor`, which retains one aligned path from each sorted index and advances only after its caller consumes that path. Push planning behavior and its external cursor shape remain unchanged.

This gives pull and push one bounded, resumable index comparison without coupling the processor to either operation.

Stacked on #468.

## Testing

- `../vendor/bin/phpunit Import/PushPlanTest.php`
- PHPCS counts unchanged for modified files; the new processor has no violations

### DIFF
--- a/packages/reprint-client/src/lib/index/class-file-index-diff-processor.php
+++ b/packages/reprint-client/src/lib/index/class-file-index-diff-processor.php
@@ -1,0 +1,264 @@
+<?php
+
+use function Reprint\Importer\decode_local_index_entry;
+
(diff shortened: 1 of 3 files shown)


adamziel:
The stack reorder made this rewritten head an ancestor of its former #468 base, so GitHub recorded this PR as merged even though the commit is not in trunk. #541 replaces it against trunk; #468 now follows #541.

PR #538, opened 2026-08-08
[PHP] Use FileIndexDiffProcessor for files-pull

adamziel:
Files-pull now uses `FileIndexDiffProcessor` instead of keeping a second index merge in `ImportClient`.

## Background

Pull and push both compare path-sorted file indexes. Push planning already uses `FileIndexDiffProcessor`. Pull still opened two `RemoteIndexReader` objects and merged their entries itself.

## This change

Pull passes the existing remote-index decoder to the shared processor. Remote absolute-path checks and field defaults stay in `RemoteIndexReader`.

The processor now requires one index entry for every physical line. Blank or invalid records fail instead of disappearing from the diff. `RemoteIndexReader::next_entry()` keeps its old blank-line handling for standalone reads, but its line decoder is strict when the diff processor uses it.

A partial diff stores one boundary:

```json
{
  "index_diff_cursor": {
    "old_index_byte_offset": 512,
    "new_index_byte_offset": 1024,
    "preceding_new_index_entry_path_b64": "L3dwLWNvbnRlbnQ="
  },
  "fetch_list_byte_offset": 256,
  "pull_index_wal_byte_offset": 128
}
```

The processor cursor and both output offsets move together. On resume, pull discards output bytes after the saved offsets and reads the next path again. The retained remote index stays unchanged during the diff. Pull saves `fetch` as the next stage before the WAL rewrites that index.

This is the first PR in the rebuilt files-pull stack. It does not add or change a CLI option.

## Testing

```sh
cd tests
../vendor/bin/phpunit Import/FileIndexDiffProcessorTest.php Import/RemoteIndexReaderTest.php Import/PullStateTest.php Import/FilesPullStateTest.php Import/OnlyFilesPathPrefixDiffTest.php Import/PullIndexWalTest.php Import/FilesPullLocalIndexTest.php Import/FilesPullDiffResumeTest.php
cd ..
vendor/bin/phpstan analyze --memory-limit=1G
vendor/bin/phpcs --standard=PHPCompatibility --runtime-set testVersion 7.4- -ps packages/reprint-client/src
git diff --check
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -699,10 +699,14 @@ fresh.
     "updated_at": "2025-01-15T10:30:00Z"
   },
   "diff": {
-    // Byte offset into the next remote index.
-    "next_remote_index_byte_offset": 1024,
-    // Last remote index entry path consumed before that byte offset.
-    "last_consumed_remote_index_entry_path": "base64..."
+    "index_diff_cursor": {
+      "old_index_byte_offset": 512,
+      "new_index_byte_offset": 1024,
+      "preceding_new_index_entry_path_b64": "base64..."
+    },
+    // Output bytes covered by the index-diff cursor.
+    "fetch_list_byte_offset": 256,
+    "pull_index_wal_byte_offset": 128
   },
   "index": {
     "cursor": "..."               // file_index cursor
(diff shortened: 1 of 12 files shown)


adamziel:
GitHub closed this PR when its temporary base branch was deleted after #573 merged. The same branch is rebased onto trunk in #574.

PR #540, opened 2026-08-10
`--target-*` options for the `apply-runtime` command

fredrikekelund:
`apply-runtime` now takes the same `--target-*` options as `db-apply`, so a caller that keeps its own database can generate a working runtime without running `db-apply` at all.

## Background

Studio supports selective sync, which makes the database optional and lets the user select all/some/no files. To make this work with Reprint, Studio runs `pull-files`, `pull-db`, `flat-docroot`, and `apply-runtime` directly (instead of a single `pull` command with something like a `--skip-database` option).

Because the database is optional during selective syncs, Studio does not always run the `pull-db` Reprint command. This is a problem because `pull-files` -> `flat-docroot` -> `apply-runtime` yields an incorrect database config that has no SQLite target. If this is the first time Reprint pulls the remote site (and there's no preexisting DB config in Reprint state), the Studio site will be borked.

This PR solves the problem by adding the `--target-*` flags (like `--target-engine`) from `pull-db` to the `apply-runtime` command, too. 

## Implementation

An important distinction between how `db-apply` and `apply-runtime` handle the `--target-*` options is that `db-apply` saves the parameters back to state on disk, whereas `apply-runtime` does not. This essentially preserves existing Reprint behavior, where `apply-runtime` only reads database state.

When `--target-*` options are supplied, Reprint now uses them first. Missing details can still come from the database target recorded in state where that is safe. But a caller must explicitly provide `--target-engine` first, and a stated engine does not mix with saved details for another engine.

To support this setup, there's a new `DatabaseTargetResolver` class that resolves the correct values for the different flags between the CLI params and recorded state. That class returns a `DatabaseTarget` model instance that holds the actual values as props.

## Testing

I'll publish a Studio PR shortly that depends on this change (i.e., by removing the existing Studio hacks that directly overwrite the database state in Reprint's state file). Testing that PR will be a good way of manually verifying this change.


adamziel (commented):
Thank you @fredrikekelund! I'm on board with what this PR wants to achieve and the bulk of the diff. One thing I'd change, though: could we avoid introducing new `Database*` abstractions and keep it simple in the huge import class? We'll be chunking out some cli handling eventually and I'm happy to discuss the right abstractions, but we don't have to block these new CLI flags on that.

fredrikekelund:
Thanks, @adamziel! Sensible feedback 👍

I've collapsed all of the resolver and getter logic into a single `resolve_apply_runtime_database_target()` method in `import.php`. This is good for another look

PR #556, opened 2026-08-11
[PHP] Render cautious target paths with protocol slash spelling

adamziel:
Cautious opaque-text URL rewriting can now apply a restricted path from the destination URL while preserving the slash spelling already used by the text.

For this mapping:

```php
[
    'https://source.example/media' => 'https://destination.example/assets',
]
```

these inputs become:

```text
https://source.example/media/logo.png
https://destination.example/assets/logo.png

https:\/\/source.example\/media\/logo.png
https:\/\/destination.example\/assets\/logo.png
```

The candidate pattern captures the first slash in the matched `protocol://`. Each slash in `/assets` copies those exact bytes. The same capture works for literal, one-backslash, and three-backslash spellings. Protocol changes and target paths can be applied to the same match.

A target path requires an absolute URL match because a scheme-less authority has no protocol slash to copy. Scheme-less rewriting remains available for mappings without a target path. Target path components accept ASCII letters, digits, hyphens, and underscores; nested components are allowed. Other target URL parts keep their existing validation.

## Testing

```
cd tests
../vendor/bin/phpunit UrlRewriting
```

PHPStan and PHPCS also pass. `StructuredDataUrlRewriterTest.php` retains its one existing class-brace PHPCS error.

### DIFF
--- a/tests/UrlRewriting/CautiousURLBaseProcessorInTextWithMixedUnknownEscapeRulesTest.php
+++ b/tests/UrlRewriting/CautiousURLBaseProcessorInTextWithMixedUnknownEscapeRulesTest.php
@@ -216,6 +216,41 @@ class CautiousURLBaseProcessorInTextWithMixedUnknownEscapeRulesTest extends Test
                 'https://site.com/source.com/media?next=https://destination.com/media/logo.png',
                 ['https://source.com' => 'https://destination.com'],
             ],
+            'literal target path replaces a configured source path' => [
+                'https://source.example/media/logo.png',
+                'https://destination.example/assets/logo.png',
+                ['https://source.example/media' => 'https://destination.example/assets'],
+            ],
+            'target path accepts hyphen underscore and nested components' => [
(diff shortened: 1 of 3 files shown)


tests/UrlRewriting/CautiousURLBaseProcessorInTextWithMixedUnknownEscapeRulesTest.php
```diff
@@ -230,10 +270,35 @@ public static function unsupported_cases(): array
...
+            ],
+            'scheme-less authority without a separator stays unchanged' => [
+                'source.example',
+                'source.example',
+                ['https://source.example' => 'https://destination.example/assets'],
+            ],
+            'scheme-less authority with a path stays unchanged' => [
+                'source.example\\/logo.png',
+                'source.example\\/logo.png',
+                ['https://source.example' => 'https://destination.example/assets'],
```
adamziel, line 301: We can solve most or all of these problems, let's keep exploring in a follow-up PR

PR #557, opened 2026-08-12
Add `--on-flatten-to-conflict=error|replace|adopt` flag to supersede `--force`

fredrikekelund:
This PR addresses the same fundamental issue as https://github.com/WordPress/reprint/pull/383, but takes a new approach: instead of keeping the pre-existing files and directories in the `flatten_to` dir, they are now moved into the fs-root.

---

This PR adds an `--on-flatten-to-conflict=error|replace|adopt` flag to the `flat-docroot` and `pull` commands. `error` is still the default. The old `--force` flag is now an alias for `replace`, and both behave exactly as before.

`flat-docroot --on-flatten-to-conflict=adopt` keeps the plugins, themes, and uploads that only the `flatten_to` dir had, instead of deleting them.

## Background

In Studio, there's always a local WordPress site that we use Reprint to pull into. That local site can be freshly minted and contain nothing but the default WordPress placeholder content, or it can contain a bunch of existing plugins/themes/uploads. When it does have pre-existing content, we want that content to be preserved on the first Reprint pull.

`flat-docroot --force` would wipe it, which is exactly what we don't want. 

https://github.com/WordPress/reprint/pull/383 is not a good solution to this problem, as anything that lives only in the `flatten_to` dir is invisible to Reprint when pushing back to the site (we will soon also implement Reprint-based pushes in Studio). 

## Exact behavior

- Any immediate descendant of `wp-content` in flatten-to that **is not** present in fs-root **is moved** wholesale. Example: a `debug.log` file, a `plugins` dir that exists only in flatten-to, or any custom directory.
- Any immediate descendant of `wp-content` in flatten-to that **is** present in fs-root is **not moved**. 
- In the `plugins`, `themes`, and `mu-plugins` directories, immediate descendants **are moved** from flatten-to into fs-root only if the same name does not exist in fs-root. Example: flatten-to has a `woocommerce` plugin, and so does fs-root. In this case, fs-root wins, which prevents problems if the plugin/theme versions do not match.
- In the `uploads` directory, all descendants (not only immediate ones) that exist in flatten-to but not in fs-root **are moved** into fs-root.
- Symlinks that are moved from flatten-to into fs-root are recreated if the target path is relative, so that they keep pointing to the right path.

## Testing instructions

TBD. I'm looking to spin up a Studio PR that lets us test this through the `pull-reprint` command.

### DIFF
--- a/packages/reprint-client/src/import.php
+++ b/packages/reprint-client/src/import.php
@@ -173,6 +173,13 @@ class ImportClient
      */
     public const FLATTEN_TO_CONFLICT_MODES = ['error', 'replace', 'adopt'];
 
+    /**
+     * Suffix for the path a cross-filesystem adopt copies into before renaming
+     * it to its real name. Derived from the destination rather than random, so
+     * a later run clears what an interrupted one left behind.
+     */
+    private const FLATTEN_ADOPT_STAGING_SUFFIX = '.reprint-adopt-incomplete';
+
     private const SAVE_STATE_EVERY_N_CHUNKS = 50;
     private const STATE_PATH_ENCODING_PREFIX = "base64:";
     private const SQLITE_PREPARED_INSERT_CACHE_MAX = 128;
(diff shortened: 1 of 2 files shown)


fredrikekelund:
Closing in favor of https://github.com/WordPress/reprint/pull/571

PR #560, opened 2026-08-12
[PHP] Classify CSS and JSON script text

adamziel:
This stacked draft recognizes CSS in style tags and JSON in script tags declared as application/ld+json or application/json. Their raw modifiable text now uses the cautious base replacement, preserving its existing quote and escape spelling.\n\nStyle attributes keep their existing structured CSS path; this change does not broaden CSS escape decoding or arbitrary attribute processing.


adamziel:
Folded this classification change into #558, which now stacks directly on #559.

PR #564, opened 2026-08-12
Accept a file path in `pull-files --only`

### LINKED ISSUE #539: `pull-files --only` should support file paths, not just directories
Running `pull-files --only PATH_TO_FILE` fails like so:

```
{
  "status": "error",
  "command": "pull-files",
  "failed_stage": "files-pull",
  "completed_stages": [
    "preflight"
  ],
  "error_code": "SERVER_ERROR",
  "message": "Pull failed at files-pull: The remote server crashed because list_dir was a file path rather than an accessible directory: PATH_TO_FILE"
}
```

@adamziel and I discussed this on Slack and agreed that `pull-files --only` should support file paths and not just directory paths.

gavande1:
## Related issues

- Fixes #539

## Proposed Changes

`pull-files --only PATH` used to accept only directory paths. Passing a single file crashed the export server with `list_dir does not exist or is not accessible`, and the whole pull failed. 

This PR changes that so the exporter now accepts a file, directory, or symlink for the `directory` param sent to the `file_index` endpoint. The endpoint tracks the requested path and resolves symlinks (if `--follow-symlinks` is enabled). 

Aside from making `pull-files --only PATH_TO_FILE` work, this PR also makes `pull-files --only PATH_TO_SYMLINKED_DIR` work better in the sense that Reprint will now return an index containing `symlink_dir -> target` (this was previously implemented separately in https://github.com/WordPress/reprint/pull/572, but I folded it into this PR).

## Testing Instructions

This is tricky to test on a WordPress.com Atomic site, unfortunately, as the wpcomsh Reprint exporter code loads before the standalone Reprint exporter plugin most of the time. Ideally, try this on a non-wpcomsh site.

1. Build the client PHAR and exporter plugin:

   ```sh
   composer run build:phar
   composer reinstall wp-php-toolkit/reprint-server --working-dir=reprint-server-wp
   composer run build:server-plugin
   ```

2. Install `reprint-exporter-wp.zip` on the remote WordPress site. Set a shared secret in the plugin settings.
3. Verify the connection:

  ```
   php reprint.phar preflight 'https://example.com/?reprint-api' \
     --secret=TOKEN \
     --state-dir=/tmp/reprint-state \
     --fs-root=/tmp/reprint-files
   ```

4. Select a regular file path reported by preflight and run a scoped pull:

  ```
   php reprint.phar pull-files 'https://example.com/?reprint-api' \
     --secret=TOKEN \
     --state-dir=/tmp/reprint-state \
     --fs-root=/tmp/reprint-files \
     --only=/absolute/remote/path/to/file.php \
     --progress=jsonl
   ```

5. Confirm the command reports one indexed/downloaded file and that only the selected file exists below `--fs-root`.


gavande1:
cc @fredrikekelund @adamziel I am looking for early review.

adamziel:
Could we test this repro: pull `--only /site/foo.php`, delete the remote file, then run the same pull again? I think `resolve_directories()` fails before missing-file handling runs, leaving the local copy behind.

gavande1:
Hey @adamziel, I am able to reproduce the issue on this branch. The `resolve_directories()` throws before `endpoint_file_index()` builds an index, so the diff never runs and the local copy stays behind.

The same thing happens on trunk with a directory root, so it looks pre-existing rather than new here.

What we can do here is skip vanished entries in `resolve_directories()` instead of throwing. That alone is not enough though: with a single `--only FILE` root that is gone, the skip empties `$directories` and `"No valid directories specified"` throws somewhere in `export.php`. So we would also need an empty root set to mean "index nothing", which changes behavior for every caller. Do you prefer it as a follow-up, or in this PR?

Also, without `--abort` the re-run short-circuits on `already_complete` and never contacts the server, so the stale copy survives with no error at all.

packages/reprint-server/src/class-file-index-processor.php
```diff
@@ -99,12 +102,9 @@ public static function start(
         bool $include_caches,
         string $storage_path
     ): self {
-        // Anchor traversal to a real directory. All later comparisons use
-        // canonical paths so configured roots and followed links share one
-        // path namespace.
-        clearstatcache(true, $index_directory);
```
adamziel, line 107: Should we retain the `clearstatcache` call?

adamziel: Ah, it's in the `canonical_root_path` function now. 👀

packages/reprint-server/src/utils.php
```diff
@@ -393,6 +394,35 @@ function path_is_same_as_or_descendant_of($path, $ancestor): bool
     return $path === $ancestor || str_starts_with($path, $ancestor . "/");
 }
 
+/**
+ * Canonicalizes one configured root path.
+ *
+ * A non-directory keeps its final component, so a named symlink stays itself.
```
adamziel: What's "a named symlink"? Let's explain this here with an example.

packages/reprint-server/src/class-file-index-processor.php
```diff
@@ -765,6 +748,120 @@ private static function position_after_name(array $directory_names, string $afte
...
+            "path" => $path,
+            "ctime" => (int) ( isset($stat["ctime"]) ? $stat["ctime"] : 0 ),
+            "size" => $type === "file" || $type === "link" ? (int) ( isset($stat["size"]) ? $stat["size"] : 0 ) : 0,
+            "type" => $type,
+        ];
+        if ($link_target !== null) {
+            $item["target"] = $link_target;
+        }
+        if ($type === "dir") {
+            // Physical emptiness, not post-exclusion: else push deletes the parent.
```
adamziel:
```suggestion
            // Actual empty directory, not a directory with all its children excluded
            // from the synchronization.
```

packages/reprint-server/src/class-file-index-processor.php
```diff
@@ -765,6 +748,120 @@ private static function position_after_name(array $directory_names, string $afte
...
+                        break;
+                    }
+                    if ($directory_entry !== "." && $directory_entry !== "..") {
+                        $item["empty"] = false;
+                        break;
+                    }
+                }
+                closedir($directory_handle);
+            }
+            // An absent "empty" means unknown, so push must not infer deletions.
```
adamziel:
```suggestion
            // If opendir() fails, do not set the "empty" key. Let pull report the directory
            // error, and let push abort before planning deletions.
```

packages/reprint-server/src/class-file-index-processor.php
```diff
@@ -765,6 +748,120 @@ private static function position_after_name(array $directory_names, string $afte
...
+        }
+
+        // One step: the cursor cannot stop between a link and its intermediates.
```
adamziel:
```suggestion
        // Intermediate links and the inspected path belong to the same step
        // because the cursor cannot stop between them without losing one.
```

packages/reprint-server/src/class-file-index-processor.php
```diff
@@ -765,6 +748,120 @@ private static function position_after_name(array $directory_names, string $afte
...
+        $entries = $intermediate_symlinks;
+        // A descendant implies its ancestors, so only empty directories need a row.
```
adamziel:
```suggestion
	// We skip non-empty directories—they're implied by their descendants.
	// For example, a row for "/a/file" already implies that "/a" exists.
	// Only emit directory entries when they're empty or uninspectable because
	// no descendant can imply their existence.
```

packages/reprint-server/src/class-file-index-processor.php
```diff
@@ -284,6 +320,12 @@ public function next_index_step(): bool
             return true;
         }
 
+        // One per step, before traversal, so each step stays bounded.
```
adamziel:
```suggestion
		// `--only` may select files such as wp-config.php. Index one such file now;
		// after all selected files are indexed, continue with directory walking below.
```

packages/reprint-server/src/class-file-index-processor.php
```diff
@@ -61,6 +61,9 @@ final class FileIndexProcessor {
     /** @var array[] Intermediate symlinks emitted before a new traversal begins. */
     private $initial_index_entries;
 
+    /** @var string[] Named paths still to inspect, one per step, before traversal. */
+    private $pending_path_roots = [];
```
adamziel: Let's clarify the purpose of this:

```suggestion
	/** @var string[] Requested non-directory paths still to index. */
	private $pending_non_directory_paths = [];
```

packages/reprint-server/src/class-file-index-processor.php
```diff
@@ -765,6 +748,120 @@ private static function position_after_name(array $directory_names, string $afte
         return $low;
     }
 
+    /**
+     * Inspects one named path, applying the omissions traversal applies.
+     */
+    private function index_next_path_root(): void
+    {
```
adamziel, line 768:
```suggestion
	/**
	 * Indexes the next requested non-directory path using the same exclusions as directory traversal.
	 */
	private function index_next_non_directory_path(): void
	{
```

packages/reprint-server/src/class-file-index-processor.php
```diff
@@ -765,6 +748,120 @@ private static function position_after_name(array $directory_names, string $afte
...
+        clearstatcache(true, $path_root);
+        $stat = @lstat($path_root);
+        if ($stat === false) {
+            $this->step_status = self::STATUS_PATH_UNAVAILABLE;
+            return;
+        }
+
+        $entries = [];
+        if ($this->follow_symlinks) {
+            // dirname() so a symlinked file is not repeated as an intermediate entry.
```
adamziel:
```suggestion
            // Record any symlinked descendants. The `index_entries_for_path` call
            // below ignores them and only considers one specific path.
```

adamziel: Actually that's not true, it does not ignore symlinks 👀 :

```php
            $resolved_symlink = self::resolve_symlink_target($path);
            $link_target = $resolved_symlink["target"];
            if ($follow_symlinks) {
                $intermediate_symlinks = $resolved_symlink["intermediates"];
            }
```

adamziel: @gavande1  Suppose `/current` is a symlink to `/releases/42`, and the user requests `--only /current/wp-config.php`. This code receives `/releases/42/wp-config.php` because the parent path was already resolved. It can no longer find the `/current` symlink, so pull creates `<fs-root>/releases/42/wp-config.php` but not `<fs-root>/current`. Let's preserve the requested path when finding parent symlinks and add a test for this case.

adamziel (changes requested):
> Do you prefer it as a follow-up, or in this PR?

A follow-up sounds fine! That being said, I found [one blocker](https://github.com/WordPress/reprint/pull/564#discussion_r3783347459)

fredrikekelund:
Rahul is AFK this week, so I'm taking over this PR. Thanks for the initial review, @adamziel 👍

fredrikekelund:
@adamziel, I've refactored this pretty drastically. The big thing is that `FileIndexProcessor` now reasons about arrays containing `requested_path`, `resolved_path`, and `type` for every root entry. This is much more expressive and makes it easier to ensure the requested path, its resolved parent directories, and the resolved target are added to the index.

I opened https://github.com/WordPress/reprint/pull/572 on Thursday to fix a case where `pull-files --only wp-content/plugins/akismet` would transfer `akismet` as a directory instead of a symlink and a resolved target. I moved that functionality into this PR today. It's closely related to the first purpose of this PR.

adamziel:
@fredrikekelund thank you! The tests fail here and I've merged a bunch of changes in the meantime that caused conflicts. Let me see if I can get it rebased and all green.

packages/reprint-client/src/import.php
```diff
@@ -7336,10 +7346,12 @@ private function fetch_file_batch(
         }
 
         $params = $this->get_tuned_params("file_fetch");
-        // Always send directory[] – see comment in fetch_next_remote_index().
-        $export_dirs = $this->get_export_directories();
-        if (!empty($export_dirs)) {
-            $params["directory"] = $export_dirs;
+        // file_fetch retains its directory-only server contract. --only may
```
adamziel: This comment sounds like it's meant for this PR's reviewer. "Retaining a contract" won't make much sense once this code is merged. Let's rewrite it or remove it.

packages/reprint-client/src/import.php
```diff
@@ -7888,12 +7904,20 @@ private function compare_remote_indexes_and_build_fetch_list(): bool
                     $remote_absolute_path = $index_diff->get_path();
                     $transition = $index_diff->get_path_transition();
                     if ($transition === "deleted") {
+                        $remote_path_type =
+                            $index_diff->get_path_type_in_old_index();
+                        if ($remote_path_type === null) {
+                            throw new LogicException(
```
adamziel, line 7919: This is here just because the return type is nullable, correct? As in, there's no logical way for this to happen?

packages/reprint-client/src/import.php
```diff
@@ -9784,19 +9817,65 @@ private function resolve_remote_paths(
...
+    /**
+     * Returns selected roots that the prior remote index confirms as tracked.
+     *
+     * The server may represent only these roots as an empty selected result
+     * when it can confirm their current absence. A newly typed missing path is
+     * still rejected at the endpoint.
+     *
+     * @return string[] Selected roots present in the prior remote index.
+     */
+    private function previously_indexed_selected_roots(): array
```
adamziel: Something about the side effects nature of this method feels off but I can't articulate it. Note to self: revisit this on another day. Not a blocker for merging.

packages/reprint-client/src/import.php
```diff
@@ -9824,7 +9904,12 @@ private function is_selected_for_pulling(
                     $included_path_prefix
                 );
                 if ($remainder === "") {
-                    return false;
+                    // Directory roots have no row in a freshly indexed tree,
+                    // so their old row must survive a scoped delta. Named
```
adamziel: What's a "scoped delta"?

packages/reprint-client/src/lib/push/class-push-plan.php
```diff
@@ -156,9 +156,14 @@ public static function start(
         if (!is_resource($plan->fresh_local_index_handle)) {
             throw new RuntimeException("Failed to open the fresh local index: {$plan->fresh_local_index_file}");
         }
+        $filesystem_root_record = [
```
adamziel, line 161: Should we declare a new class to represent this data type? Not a blocker.

packages/reprint-server/src/export.php
```diff
@@ -1326,6 +1328,166 @@ function resolve_directories(array $config): array
...
+            "resolved_path" => $resolved_path,
+            "type" => $type,
+        ];
+    }
+
+    return $roots;
+}
+
+/**
+ * Resolves the root scheduled first for a file-index request.
```
adamziel: Could we add an example? I'm struggling to understand this comment and an example would help me.

packages/reprint-server/src/export.php
```diff
@@ -1326,6 +1328,166 @@ function resolve_directories(array $config): array
...
+
+    return [
+        "requested_path" => $requested_path,
+        "resolved_path" => $resolved_path,
+        "type" => "directory",
+    ];
+}
+
+/** Returns whether the parent can confirm that a selected name is absent. */
+function file_index_root_is_confirmed_absent(string $requested_path): bool
```
adamziel: This is s weird function. Couldn't we just call stat?

packages/reprint-server/src/export.php
```diff
@@ -1326,6 +1328,166 @@ function resolve_directories(array $config): array
...
+{
+    $parent = dirname($requested_path);
+    $names = @scandir($parent, SCANDIR_SORT_NONE);
+    return is_array($names) && !in_array(basename($requested_path), $names, true);
+}
+
+/** Returns the first symlink in a requested root's parent path. */
+function file_index_parent_symlink(string $requested_path): ?array
```
adamziel, line 1424: There may be some existing helper that helps with this or a similar problem. I'm not 100% sure though.

tests/e2e/tests/import-64-only-symlinked-directory-root.test.js
```diff
@@ -0,0 +1,66 @@
...
+            secret: getSiteSecret(site),
+            extraArgs: ['--only', join(siteDir, 'wp-content', 'plugins', 'akismet')],
+        });
+        assert.equal(
+            result.exitCode, 0,
+            `Expected exit 0\nstderr: ${result.stderr}\nstdout: ${result.stdout}`,
+        );
+    });
+
+    it('recreates the selected symlink with its original target spelling', () => {
```
adamziel, line 52: This might bite when migrating from macos where the filesystem is case insensitive. It's okay at the moment, it's not a supported scenario anyway.

adamziel (approved):
Looking good! I left some minor notes that we could address here or in a follow up pr. I see no blockers. Thank you!

packages/reprint-client/src/lib/push/class-push-plan.php
fredrikekelund: I considered this, mainly for type safety. I decided not to because there are guidelines in `AGENTS.md` that say this:

> Inline single-use helpers; no wrapper classes that only rename a concept. A "transport" callable, a plan-then-send request object, and a processor that materialized a file list were all deleted from this repo. A helper earns its name at two or more callers.

For type safety, we use a PHPDoc `FileIndexRoot` type that PHPStan checks in the `composer analyze` script. We can revisit this if we want, but the current approach is OK, IMO.

packages/reprint-server/src/export.php
fredrikekelund: Good question. We call `lstat()` on the path before calling this function, but `lstat() === false` doesn't guarantee that the path was deleted. It could also mean `lstat()` failed. This function returns true only if we can guarantee the path is no longer present. It's a sensible addition, because the final effect of this logic is that the client deletes the path in question from its local file system.

packages/reprint-server/src/export.php
fredrikekelund: I double-checked. My agent identified `FileIndexProcessor::find_parent_symlinks()` as a similar candidate, but they're not similar enough that it's clear-cut how to harmonize them. We could refactor this in the future, or leave as is.

packages/reprint-client/src/import.php
fredrikekelund: Thanks for calling this out. I rewrote this entire comment

packages/reprint-server/src/export.php
fredrikekelund: Agreed that this wasn't really understandable. I rewrote the whole comments for both this function and `resolve_file_index_roots()`

packages/reprint-client/src/import.php
fredrikekelund: Makes sense to revisit 👍

packages/reprint-server/src/export.php
adamziel: Interesting! `scandir` still worries me—it could be expensive for parents with many directories in them. I've mapped some scenarios where the two would disagree:

Situation | lstat() | scandir(parent)
-- | -- | --
Parent is readable but lacks execute/search permission | Fails with EACCES | May list the name
Entry was renamed/unlinked concurrently | May fail with ENOENT | May show old or new snapshot
NFS/FUSE/network filesystem inconsistency or I/O error | May fail transiently | May still list a cached directory entry
32-bit PHP cannot represent size/inode metadata | May fail with EOVERFLOW | Lists the name
PHP stream wrapper lacks stat support / open_basedir blocks it | Fails | Wrapper-dependent

Only the first one seems relevant to us. It seems like:

* `readdir()` requires `x` permissions on every parent in the path and `r` permissions on the probed directory
*  `lstat` requires x permission on every directory in the path

So something along these lines:

```
r-- directory:
readdir() can reveal "file.txt"
lstat("directory/file.txt") fails with EACCES

--x directory:
readdir() cannot list filenames
lstat("directory/file.txt") succeeds if you already know the name
```

Since the codebase amply assume that `lstat()` is reliable, I'd keep using it in this PR. We may need to revisit it one day, but I wouldn't bundle this change in here. How does that sound?

fredrikekelund: Came back to this today. I ended up going with just `lstat()` here, mainly because the `scandir()` solution has a much more obvious flaw: `file_index_root_is_confirmed_absent()` would return false if the parent directory were removed.

fredrikekelund:
I pushed fixes for two bugs that were identified in a final agentic review. The diff is relatively small.

1. The `ImportClient::derive_remote_deletion_root_from_sparse_index()` method was not aware of which directories the current pull had asked to index. For example, if I run `pull-files --only /var/www/html/wp-config.php` and the file is gone from the remote, then the new remote index ends up empty. `ImportClient::derive_remote_deletion_root_from_sparse_index()` would previously return `/var` in this case, which led to everything under `<fs-root>/var` being removed on the local machine, including paths the run never asked about. Fixed it by passing the result of `ImportClient::get_export_directories()` to that method and using that list to check if the current pull is "scoped" to a limited number of paths, and if so, whether the current path is encompassed by that list.
2. The `ImportClient::previously_indexed_selected_roots()` method looked for an exact path match on directory paths from the previous remote index. The problem is that such a path will never exist, because Reprint uses a "sparse" index, where directories are implicit based on their descendant file paths. Fixed with the help of `path_is_same_as_or_descendant_of()`.

Lastly, I also revisited the `ImportClient::previously_indexed_selected_roots()` naming (and the docblock comment). The method is now called `get_selected_paths_pulled_before()`, the wire param is called `pulled_before` (not `missing_roots`), and the cache prop is called `$selected_paths_pulled_before` (not `$previously_indexed_selected_root_paths`)

PR #572, opened 2026-08-13
[Pull] Preserve selected symlinked directories selected with `--only`

fredrikekelund:
`files-pull --only=<path>` now recreates `<path>` when it is a symlink to a directory, so a scoped pull leaves the selected plugin available at the path the user requested.

For example, a site may contain:

```text
wp-content/plugins/akismet -> ../../../wordpress/plugins/akismet/5.7
```

Before this change, pulling only that plugin indexes and downloads `wordpress/plugins/akismet/5.7/` but leaves `wp-content/plugins/akismet` absent under `--fs-root`. After this change, the pull also recreates `wp-content/plugins/akismet` with the same relative target spelling, so the selected plugin resolves locally.

> [!NOTE]  
> This is Codex through and through. This issue was identified in https://github.com/Automattic/studio/pull/4491/#discussion_r3764696102, then verified, planned, implemented, and tested by my agent.


fredrikekelund:
Closing this in favor of https://github.com/WordPress/reprint/pull/564

PR #581, opened 2026-08-13
[PHP] Share resumable file sync patch processing

adamziel:
Adds one resumable processor for building a fresh local index and planning a patch against a saved index.

Push keeps its current path lists and progress labels. `PushPlan` now consumes operations from the shared processor instead of coordinating `FreshLocalIndexProcessor` and `FileSyncPatchPlanner` itself.

The two start methods make the patch direction explicit:

```php
// Saved index -> current local tree. Used by push.
$processor = FileSyncPatchProcessor::start_to_fresh_local_tree(
	$work_directory,
	$filesystem_root,
	$saved_local_index,
	$storage_path
);

// Current local tree -> saved index. Used by mirror pull in the next PR.
$processor = FileSyncPatchProcessor::start_from_fresh_local_tree(
	$work_directory,
	$filesystem_root,
	$saved_local_index,
	$storage_path
);
```

The processor owns the scan, sort, patch planner, and their complete cursor. Callers store that cursor unchanged and consume one operation per planning step.

This also keeps selection roots safe in JSON cursors and lets a selected subtree be deleted without crossing a nested include or exclusion boundary.

## Testing

The processor, planner, PushPlan, and real push endpoint suites cover both patch directions, interruption at every internal phase, arbitrary path bytes, exclusions, progress, and process resume.

### DIFF
--- a/docs/PUSH-TERMINOLOGY.md
+++ b/docs/PUSH-TERMINOLOGY.md
@@ -66,6 +66,9 @@ local relative path to a document-root-relative path.
   patch result. To copy only source changes, the source snapshots from the
   last sync and now are the patch base and result. The planner owns the index
   diff and the active deletion roots file.
+- A **file sync patch processor** builds a fresh local index, then runs a file
+  sync patch planner in the direction selected by its start method. Its cursor
+  owns both phases, so a caller can store it unchanged.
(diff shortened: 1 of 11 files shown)


adamziel:

PR #582, opened 2026-08-13
[PHP] Stop pull-db cleanly during database apply

adamziel:
`pull-db` now handles SIGINT and SIGTERM while it is applying the database, exits with code 2, and leaves the database-apply stage unfinished for the next process.

## Background

The wrapper kept its normal signal handler while `db-apply` was running. A signal could therefore kill the process during an active PDO query. If `db-apply` returned `partial`, the wrapper also called it again immediately in the same process.

## This change

The wrapper uses the database-apply signal policy only for that stage and restores its normal policy afterward. Pending signals are handled between SQL chunks, not during a target query.

`db-apply` runs once per wrapper process. A partial result sets exit code 2 and returns without marking the stage or the whole pull complete. The next process uses #595 to start `db.sql` again from the beginning.

This is PR 4 of 5: #590 → #591 → #595 → #582 → #607.

## Testing

The focused PHP tests cover the partial result, next-process completion, restoration after an exception, and the actual PCNTL handlers and signal mask. The MySQL E2E blocks a target query, sends SIGTERM, checks exit code 2 and the unchanged stage, then completes the pull in a new process.

### DIFF
--- a/packages/reprint-client/src/import.php
+++ b/packages/reprint-client/src/import.php
@@ -285,6 +285,9 @@ class ImportClient
     /** @var bool Set to true by SIGTERM/SIGINT handler to finish the current chunk and exit cleanly. */
     private $shutdown_requested = false;
 
+    /** @var string|null Top-level command whose signal policy is restored after a scoped stage. */
+    private $signal_handling_command;
+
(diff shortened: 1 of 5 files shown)


adamziel:

PR #584, opened 2026-08-13
[PHP] Share remote-to-local pull path mapping

adamziel:
Files pull now uses one `RemoteToLocalPathMapper` for fetched paths, rewritten symlink targets, and followed targets. This keeps the existing placement rules in one class before mirror pulls need the same rules while building a local sync plan.

The mapper keeps the current order: the longest explicit remap wins, paths in the original pull scope keep their remote path below the local filesystem root, and followed targets outside that scope go below the followed-symlinks root. Paths remain byte strings.

```php
$mapper = new RemoteToLocalPathMapper(
    "/local",
    ["/srv/site"],
    ["/srv/site/wp-content" => "/local/content"]
);

$mapper->map_path("/srv/site/index.php");
// /local/srv/site/index.php

$mapper->map_path("/srv/site/wp-content/plugin.php");
// /local/content/plugin.php
```

This is a refactor. The existing pull endpoint tests pass without changing the resulting paths.

### DIFF
--- a/packages/reprint-client/src/import.php
+++ b/packages/reprint-client/src/import.php
@@ -107,6 +107,7 @@ require_once __DIR__ . '/lib/pull/class-pull.php';
 
 // Pull index reader and the WAL for completed files-pull mutations.
 require_once __DIR__ . '/lib/pull/class-remote-index-reader.php';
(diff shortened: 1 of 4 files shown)


adamziel:

PR #585, opened 2026-08-13
[PHP] Keep pulled deletions within selected paths

adamziel:
Pulled remote deletions now stop at the active include, exclude, and remap boundaries.

For example, when `/var/www/html` is remapped to `site` and that whole remote tree disappears, the sparse index can identify `/var` as the first absent parent. Pull now stops at `/var/www/html` and removes `site` instead of mapping `/var` to an unrelated local path. An excluded descendant or a separate remap target also keeps a broader parent from being deleted.

`FileSyncPatchPlanner` and files-pull now use the same sparse-index deletion rule. When a former empty-directory entry is replaced by a descendant, pull removes the stale index entry without removing the directory.

## Testing

The focused tests cover nested includes, excluded descendants, complete remapped-tree deletion, overlapping local remap targets, and an empty directory becoming non-empty.


adamziel:

PR #587, opened 2026-08-13
[PHP] Add exact local cleanup to file sync patches

adamziel:
File sync cleanup can now remove selected indexed paths without removing skipped or excluded children.

`FileSyncPatchProcessor::start_from_fresh_local_tree()` plans exact deletions. The push direction keeps collapsing deleted subtrees to one root.

`FileSyncCleanupProcessor` removes one exact file, link, or empty directory per step. It uses `unlink()` and `rmdir()` and never walks a directory recursively. If a cache, storage, excluded, or newly created child remains, its parent remains too. After the exact removals, cleanup prunes one unchanged empty parent per step.

The cursor stores a pending removal and the local type, size, and ctime seen by the fresh scan. Cleanup checks those fields before removing the path. If they changed, it returns `restart` and leaves the new path alone.

```php
$cleanup = FileSyncCleanupProcessor::start(
    $work_directory,
    $filesystem_root,
    $patch_result_index,
    $storage_path,
    ["wp-content"],
    ["wp-content/uploads/keep"]
);

do {
    $has_next_step = $cleanup->next_step();
    $cleanup->flush_pending_output();
    save_cursor($cleanup->get_cursor());
} while ($has_next_step);

$status = $cleanup->get_status();
$cleanup->close();

if ($status === "restart") {
    restart_file_sync();
}
```

### DIFF
--- a/packages/reprint-client/src/lib/index/class-file-sync-cleanup-processor.php
+++ b/packages/reprint-client/src/lib/index/class-file-sync-cleanup-processor.php
@@ -1,0 +1,744 @@
+<?php
+
+use function Reprint\Importer\append_file_sync_path_stack_entry;
+use function Reprint\Importer\file_sync_index_path_may_change;
+use function Reprint\Importer\read_file_sync_path_stack_entry;
+use function WordPress\Filesystem\wp_join_unix_paths;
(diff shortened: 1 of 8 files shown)


adamziel:

PR #589, opened 2026-08-14
[PHP] Retain the next cautious URL match per mapping

adamziel:
Retains the next URL match already found for each cautious rewrite mapping.

A scan with several mappings previously ran every mapping regexp after every match. Mappings whose next match was later in the text searched the same bytes again on each call. The processor now keeps their relevant captures until the scan reaches them. Captures behind the current byte offset are refreshed, and exhausted mappings stay exhausted. The single-mapping path remains direct because it has no future lookahead to reuse.

For 100 matches of one source URL, the local benchmark measured:

| Mappings | Before | After |
| ---: | ---: | ---: |
| 1 | 92.8 µs | 93.2 µs |
| 2 | 342 µs | 128 µs |
| 8 | 1.85 ms | 0.20 ms |
| 32 | 7.90 ms | 0.48 ms |
| 64 | 16.16 ms | 0.85 ms |

## Testing

The URL rewriting suite passes with 364 tests and 2,502 assertions. New cases cover a future mapping retained across repeated earlier matches and a shorter overlapping mapping refreshed after the longer source wins. PHPStan and the coding-standard checks pass for the changed files.

### DIFF
--- a/packages/reprint-client/src/lib/url-rewrite/class-cautious-url-base-processor-in-text-with-mixed-unknown-escape-rules.php
+++ b/packages/reprint-client/src/lib/url-rewrite/class-cautious-url-base-processor-in-text-with-mixed-unknown-escape-rules.php
@@ -123,6 +123,22 @@ class CautiousURLBaseProcessorInTextWithMixedUnknownEscapeRules {
     private int $bytes_already_scanned = 0;
 
     /**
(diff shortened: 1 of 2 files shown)


adamziel:
Moved the proposal and benchmark to #624. Closing this implementation because the full pipeline did not show a gain which justifies the added per-mapping state and refresh rules.

PR #590, opened 2026-08-14
[PHP] Refuse to resume interrupted stdout and MySQL database pulls

adamziel:
A new `db-pull` process now refuses to continue stdout or direct MySQL output after the earlier importer stops.

## Background

The old process could save a source position without knowing what reached stdout or what MySQL kept. A later process could therefore skip SQL, send it twice, or append the rest of a dump to a different output.

## This change

`database-pull-output.json` is written before the first stdout or MySQL write. If the process stops, the next command explains what may have happened and tells the user to reset the receiving program, file, or database before running `db-pull --abort`.

Source failures still retry inside the same importer process, where the connection and unfinished SQL remain in memory. The command also rejects short stdout writes and checks every result returned by `mysqli::multi_query()`.

This is PR 1 of 5: #590 → #591 → #595 → #582 → #607.

## Testing

The focused E2E tests stop real stdout and MySQL pulls, replace normal command state with `db-index`, and check that a later process does not send more SQL. They also cover a failure in the second statement of one MySQL group.


adamziel:

PR #591, opened 2026-08-14
[PHP] Resume db.sql downloads from the last saved file size

adamziel:
File-backed `db-pull` can now continue from the last file size and source position Reprint saved.

## Background

Exit code 2 left `db-pull` marked `partial`, but a later process only recognized `in_progress` and started over. A hard stop could also leave extra bytes at the end of `db.sql`, or leave the file shorter than the saved source position.

## This change

Reprint flushes `db.sql` before saving its size with the matching source position. A later process removes any bytes written after that saved size, then continues the download. It refuses a missing or shorter file and refuses to change the output mode halfway through the download.

Every request asks the server for SQL which drops and recreates each dumped table. After the server finishes, `database-dump.json` stores the SHA-256 of the complete `db.sql`. `db-apply` will not read a file which is still downloading.

This is PR 2 of 5: #590 → #591 → #595 → #582 → #607.

## Testing

The focused tests cover exit code 2 during database indexing, process death before and after the first saved file size, missing and shortened `db.sql`, an output-mode change, and an attempt to apply a download which has not finished.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -168,7 +168,7 @@ Entries in `paths_to_remove` under `wp-content/plugins/` also trigger automatic 
 
 ### SQL Streaming Source Retries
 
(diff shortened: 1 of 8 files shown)


adamziel:

PR #592, opened 2026-08-14
[PHP] Share file-index path inspection

adamziel:
File indexing and files-pull journal updates now derive path type, size, and ctime from one stat-record decoder.

Both paths were decoding the same Unix mode bits. There is no reason to maintain two answers for whether a path is a file, directory, link, or unsupported type. `read_file_index_entry_from_stat()` now owns that translation and the size normalization used by file indexes.

`FileIndexProcessor` still owns symlink targets and physical directory emptiness. `PullIndexJournal` still writes only the fields carried by its WAL. This changes no index format or pull behavior.

Tested with `FileIndexProcessorTest`, `PullIndexWalTest`, and `FreshLocalIndexProcessorTest`, plus the PHP 7.2 server and PHP 7.4 client compatibility scans.


adamziel:

PR #593, opened 2026-08-14
[PHP] Report remote index read failures

adamziel:
`RemoteIndexReader` now distinguishes a failed read from the end of an index.

`fgets()` returns `false` for both cases. Returning `null` without checking `feof()` lets a caller accept a partial index as complete. The reader now returns `null` only at EOF and throws a pointed exception when the stream failed before EOF.

The test uses a stream wrapper whose read fails while `stream_eof()` remains false. The existing file-backed reader tests continue to cover normal completion and resume.


adamziel:

PR #594, opened 2026-08-14
[PHP] Read remote directory shape from index entries

adamziel:
`RemoteIndexReader` now returns the optional `empty` value recorded for remote directories.

The exporter writes this field when it has scanned an empty directory. Dropping it leaves downstream index consumers unable to distinguish an explicit empty directory from an ordinary directory whose descendants imply its presence. Retained remote indexes rebuilt by `PullIndexJournal` may omit the field, so it remains optional.

The decoder accepts a boolean `empty` value only on directory entries. Tests cover the returned value, non-boolean input, and file or link entries which incorrectly carry it.


adamziel:

PR #595, opened 2026-08-14
[PHP] Restart interrupted db-apply from the beginning

adamziel:
A stopped `db-apply` now starts an unchanged, complete `db.sql` from the beginning on a new target connection.

## Background

The old code continued after its saved statement number. That statement number did not say which MySQL changes were committed. The new connection also skipped the dump header which sets SQL mode, foreign-key checks, unique checks, and autocommit.

## This change

Before the first target query, `database-apply.json` stores the SHA-256 of `db.sql`, the target database, URL replacements, table prefix, and host-specific plugin removals. A later process starts again only when those values still match. If they do not, the error explains that the target may have changed and tells the user to restore or reset it before aborting the apply.

Every restart opens a new connection and reads `db.sql` from the beginning, so the header runs again. MySQL applies also hold a lock in the target database. A new process waits if the old target session is still running a query.

Completion is saved before normal command state. If the process stops between those two writes, the next process repairs the command state without applying SQL again.

This is PR 3 of 5: #590 → #591 → #595 → #582 → #607.

## Testing

The focused tests cover orderly stops, process death, changed SQL, changed targets and options, connection failure, the final state-write failure, and a target session which remains alive after the PHP process dies. The MySQL E2E also checks that SQL mode and foreign-key settings run on the new connection.


adamziel:

PR #596, opened 2026-08-14
[PHP] Record remote sizes for pulled symlinks

adamziel:
Pulled symlinks now retain the remote target's byte length in the remote index.

The next remote index gets a symlink's size from remote `lstat()`. Fetch later wrote size `0` into the pull index WAL, so applying that record changed the metadata of an otherwise unchanged symlink. This records `strlen($target)` from the decoded remote target. It deliberately does not use the rewritten local target, whose length may differ after path mapping.

## Testing

The symlink pull test checks the WAL record and the existing symlink cases continue to pass.


adamziel:

PR #597, opened 2026-08-14
[PHP] Store remote-to-local path mapper configuration

adamziel:
Remote-to-local path mapping can now be stored in a processor cursor and recreated without callers unpacking its fields. Every configured path is base64-encoded in the cursor, so filesystem names remain arbitrary byte strings.

This is a prerequisite for resumable pull processors which map remote index entries after a process restart.

## Testing

The mapper suite covers identity placement, remaps, followed-symlink placement, precedence, and a JSON round trip with non-UTF-8 path bytes.

Stacked on #596.


adamziel:

PR #598, opened 2026-08-14
[PHP] Retain duplicate paths while sorting an index

adamziel:
Adds `sort_index_file_preserving_duplicate_paths()` for callers that must detect two input entries mapping to the same path. Existing `sort_index_file()` and three-argument `try_exec_sort_index_file()` calls continue to remove duplicates.

Both the system sort and bounded-memory PHP fallback keep equal paths adjacent when the new entry point is used.

## Testing

The sort suite covers the existing duplicate-removal contract and duplicate retention through both implementations.

Stacked on #597.


adamziel:

PR #599, opened 2026-08-14
[PHP] Build local-coordinate indexes for file pulls

adamziel:
Adds `PullLocalIndexProcessor`, a resumable processor which maps a remote index into local coordinates, sorts mapped entries, and merges them with the retained local index.

The processor builds a patch-result index from remote metadata for pre-fetch cleanup. After fetch, it builds a next local index from target-confirmed local metadata and checks it against one fresh local scan. A changed local tree returns `restart`; a matching tree returns `complete`.

Mapping collisions and file-or-link ancestors are rejected. Empty directories remain sparse when a selected or omitted child makes them nonempty. Each processor step advances one mapping, sort, merge, or verification boundary, and its JSON-safe cursor resumes without accepting unconfirmed output bytes.

## Testing

The processor suite covers remap reordering, overlapping mappings, sparse directories, scoped retention, fetch confirmation, restart after local changes, arbitrary path bytes, output truncation, and resume at each phase.

Stacked on #598.

### DIFF
--- a/docs/PUSH-TERMINOLOGY.md
+++ b/docs/PUSH-TERMINOLOGY.md
@@ -69,6 +69,17 @@ local relative path to a document-root-relative path.
 - A **file sync patch processor** builds a fresh local index, then runs a file
   sync patch planner in the direction selected by its start method. Its cursor
   owns both phases, so a caller can store it unchanged.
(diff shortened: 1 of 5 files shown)


adamziel:

PR #600, opened 2026-08-14
[PHP] Retain file-index selection across interrupted commands

adamziel:
Files-index and files-pull now retain their traversal selection across interrupted commands.

Before this change, omitting `--include-caches` or `--extra-directory` on resume could continue an index with a different selection. A conflicting direct or high-level pull command could also overwrite unfinished lifecycle state before command ownership was checked.

This checkpoints symlink following, cache inclusion, and the extra directory at the first durable boundary. Omitted options resume the saved selection. Explicit changes require `--abort`, and incompatible commands are rejected before state changes.

## Testing

Run the focused selection lifecycle, pull option, state, and real endpoint tests.

### DIFF
--- a/packages/reprint-client/src/import.php
+++ b/packages/reprint-client/src/import.php
@@ -917,25 +917,134 @@ class ImportClient
 
         $this->state = $this->load_state();
 
+        if (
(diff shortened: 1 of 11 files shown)


adamziel:

PR #601, opened 2026-08-14
[PHP] Resume remote-index sorting without another request

adamziel:
Remote-index sorting now has its own durable stage.

Sorting replaces the index atomically. If the process stops after that replacement but before the following state save, the saved lifecycle used to return to indexing and request the remote tree again. Resume now repeats the idempotent sort and proceeds without another file-index request.

## Testing

Interrupt files-index and files-pull after the sorted index replaces the unsorted file, then resume and confirm no additional file-index request is sent.


adamziel:

PR #602, opened 2026-08-14
[PHP] Keep target-confirmed file-index batches across resume

adamziel:
Fully written file-index batches now survive an interrupted request.

Each confirmed batch flushes the index and checkpoints its remote cursor with the matching byte offset. Resume truncates unconfirmed tail bytes and replays the saved request selection. Directory traversal errors stop at the last durable boundary instead of turning an incomplete index into apparent remote deletions.

## Testing

Interrupt the real exporter between batches, fail a batch checkpoint save, and resume from a new client. Confirm every path appears once. Make a real directory traversal fail and confirm the partial response is discarded.


adamziel:

PR #603, opened 2026-08-14
[PHP] Bind file-index cursors to canonical traversal selection

adamziel:
File-index cursors now reject traversal-selection changes, and the exporter reports the canonical roots covered by a completed traversal.

The cursor binds the canonical list directory, sorted directory selection, symlink and cache settings, and storage path. The client requires protocol v2 before starting versioned file-index work; a local abort remains available with stale preflight state. High-level pulls reject an old exporter before checkpointing preflight, so rerunning after an exporter upgrade performs preflight again.

## Testing

Resume cursors with reordered equivalent directories and reject changed traversal settings. Run a v1 endpoint, confirm no file-index request is sent or preflight checkpointed, switch the same endpoint to v2, and confirm the rerun reaches file indexing.


adamziel:

PR #604, opened 2026-08-14
[PHP] Replay completed remote-index traversals from durable ranges

adamziel:
Completed initial and followed remote-index traversals now resume without issuing the same traversal again.

Each traversal records its target-confirmed raw index range and canonical roots before its state advances. Followed-target discovery moves forward from a durable byte offset through the growing index, while disk-backed root markers prevent canonical targets from being revisited without loading the target set into memory. A followed request indexes only that target instead of walking the primary roots again.

Request failures stop the current run at the last durable boundary. Resume removes any unconfirmed index tail before discovery reads it, so buffered stale rows cannot start another traversal.

## Testing

Interrupt initial and followed traversals at request, batch, discovery, and pre-sort boundaries, then resume through the real exporter. Confirm completed requests are not repeated, empty indexes preserve the zero-byte sentinel, the first link at byte zero is consumed once, target-only requests discover nested links, and unconfirmed rows are discarded before discovery resumes.


adamziel:

PR #605, opened 2026-08-14
[PHP] Resume external sorting from durable merge boundaries

adamziel:
Adds a line-sort processor that resumes from durable run, merge, copy, and publish boundaries.

The existing external sorter performs its split and k-way merge in one call and retains one handle per run. Files-pull ownership needs to sort index-sized artifacts inside a processor lifecycle. This processor uses two deterministic framed work files, fixed-size steps, strict JSON-safe cursors, checked file operations, and atomic output replacement. A follow-up PR will use it for ownership snapshots.

## Testing

The focused tests cover multi-pass and odd-run merges, duplicate and binary keys, bounded reads and memory, replay after every cursor, work-file truncation, output publication, cleanup, malformed state, write limits, and rename failures.

### DIFF
--- a/packages/reprint-client/src/lib/class-external-sort-processor.php
+++ b/packages/reprint-client/src/lib/class-external-sort-processor.php
@@ -1,0 +1,1066 @@
+<?php
+declare(strict_types=1);
+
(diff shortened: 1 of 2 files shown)


adamziel:

PR #607, opened 2026-08-14
[PHP] Download db.sql before applying MySQL database pulls

adamziel:
`db-pull --sql-output=mysql` now downloads all of `db.sql` before it changes MySQL.

## Background

The direct MySQL path executed SQL while it arrived. If that process stopped, the saved source position could not say whether MySQL committed, rolled back, or was still running a query.

## This change

MySQL output now uses the database-only pull sequence: preflight, file-backed `db-pull`, then `db-apply`. The existing MySQL options become target options for `db-apply`, and the direct `mysqli::multi_query()` path is removed.

A stopped download continues from the last saved file size and source position. MySQL remains untouched until the download finishes. If applying the SQL stops, the next process opens a new connection and starts `db.sql` from the beginning. That runs the SQL mode, foreign-key, unique-check, and autocommit settings again.

Existing unfinished direct-MySQL work remains rejected instead of being guessed or converted.

This is PR 5 of 5: #590 → #591 → #595 → #582 → #607.

## Testing

The focused E2E stops after a file position is saved, checks that MySQL is still untouched, runs the same command again, and compares the finished target with the source. Its tables also require the SQL mode and foreign-key settings from the dump header.


adamziel:

PR #608, opened 2026-08-14
[PHP] Record completed files-pull ownership by selection

adamziel:
Scoped file pulls now retain a durable record of the remote paths covered by each completed traversal, including followed roots and intermediate symlinks.

The raw traversal ranges are consumed before index sorting into a byte-safe, disk-backed snapshot. Snapshot construction is bounded and resumable through the external sorter. A successful pull replaces only that selection's committed snapshot; abort retains a completed candidate after diff or fetch and removes incomplete artifacts.

This gives later catch-up and mirror stages an ownership boundary without loading the path set into memory.


adamziel:

PR #610, opened 2026-08-14
[PHP] Bound files-pull catch-up to completed ownership

adamziel:
Files-pull catch-up now changes only remote-index paths owned by the active selection or its earlier completed snapshots.

The current snapshot wins in overlaps. Other selections protect their paths; explicit exclusions and ordinary default-skipped descendants stay unchanged. Producer-emitted intermediate symlinks remain selectable through exact ownership. Final intermediate-symlink recreation uses the same ownership and preserve-local decisions, so excluded or preserved links are neither rewritten nor journaled. Missing directories require subtree ownership before recursive removal, so local additions under a deleted remote directory survive and appear in `files-diff`.

Ownership lookups stay disk-backed and bounded through the snapshot hash indexes.


adamziel:

PR #611, opened 2026-08-14
[PHP] Map files-pull ownership into local index paths

adamziel:
Files-pull ownership can now be evaluated in local-index coordinates after remaps and followed-symlink placement.

The local scope inverse-maps each candidate through every valid placement, forward-verifies aliases, and requires all owned aliases to permit a change. This leaves remap source holes, protected overlaps, excluded paths, and discontinuous subtrees untouched while allowing another mapping to fill a source hole.

The scope keeps one canonical mapper configuration and preserves disk-backed snapshot lookups, arbitrary-byte paths, and bounded handles. Cleanup and final mirror verification will consume it in follow-up changes.

### DIFF
--- a/packages/reprint-client/src/lib/index/class-file-sync-change-scope.php
+++ b/packages/reprint-client/src/lib/index/class-file-sync-change-scope.php
@@ -1,16 +1,19 @@
 <?php
 declare(strict_types=1);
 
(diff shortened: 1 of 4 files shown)


adamziel:

PR #612, opened 2026-08-14
[PHP] Retain pulled default-skipped local index rows

adamziel:
Files-pull now retains local-index metadata for every selected path it actually changes, including paths that normally match cache or development skips.

The pull journal already receives only successful mutations. Removing its second default-skip filter keeps local upserts and deletions aligned with their remote records while paths outside the filesystem root and the root itself remain excluded. This preserves arbitrary-byte paths and gives later cleanup and verification an accurate baseline for selected intermediate links.


adamziel:

PR #613, opened 2026-08-14
[PHP] Materialize selected skipped paths for local sync

adamziel:
Mirror reconciliation must inspect selected paths that ordinary cache-skipping traversal omits. This change materializes those paths in a bounded, resumable processor and binds the published list to the local change scope.

It also centralizes the scheduled-root traversal-skip predicate so remote selection, local mapping, and later supplemental indexing share one rule. Unsafe cache-skipping mappings fail before local mutation unless exclusions or protected ownership make the entire mapped region unreachable.


adamziel:

PR #614, opened 2026-08-14
[PHP] Index selected skipped paths without traversing caches

adamziel:
Fresh local indexing must observe selected paths that ordinary cache-skipping traversal omits without entering unrelated skipped trees. This change adds a bounded supplemental phase that reads the materialized path list and stats only those exact paths.

Unsupported selected path types remain visible to patch planning so an owned desired link can replace a local FIFO or socket, while inaccessible or escaping parents fail before an incorrect absence is recorded.


adamziel:

PR #615, opened 2026-08-14
[PHP] Apply local ownership scope while building pull indexes

adamziel:
PullLocal now maps and confirms only remote rows owned by the active selection and allowed by every local alias. This prevents an unowned remote alias from borrowing authority from another path which maps to the same local destination.

Final verification uses the selected skipped-path list instead of traversing every skipped tree, so exact owned links remain observable without touching unrelated FIFOs, sockets, or unreadable cache directories.


adamziel:

PR #616, opened 2026-08-14
[PHP] Restrict file-sync cleanup to selected local paths

adamziel:
File-sync cleanup now derives all removal authority from one local change scope. Deletes use the prior entry type, replacements use the desired entry type, and parent pruning requires subtree authority, so exclusions, protected aliases, and mapping holes stop destructive cleanup.

The initial local scan supplements only selected skipped paths instead of traversing every cache tree. Exact selected links can be removed or replaced while unrelated skipped FIFOs and unreadable directories remain untouched.


adamziel:

PR #617, opened 2026-08-14
[PHP] Make mirror the default files-pull intent

adamziel:
`pull`, `pull-files`, and `files-pull` now mirror selected files by default; `--intent=catch-up` preserves local-only changes.

Mirror saves its intent before preflight, reconciles selected local paths through ownership-scoped bounded processors, resumes from durable cursors, and rejects `preserve-local`. Catch-up keeps the existing transition-only behavior. Active and completed lifecycles retain their saved intent until abort, so a rerun cannot silently change semantics.

Tests cover direct and high-level option handling, interruption and restart boundaries, protected remap aliases, selected skipped paths, local drift restoration, selected local-only removal, and explicit catch-up preservation.


adamziel:

PR #618, opened 2026-08-14
[PHP] Resume interrupted pull-db downloads and database applies

adamziel:
An interrupted `pull-db` now continues the `db.sql` download or restarts the database apply instead of asking the user to discard work that Reprint can reuse.

File downloads save the file size together with the source position, after the bytes have been flushed. A later process removes any later, unfinished bytes and continues from that saved position.

Database apply starts `db.sql` from the beginning after an interruption. The new connection therefore runs the dump header again, including `SQL_MODE`, foreign-key checks, and autocommit settings. MySQL output uses this same download-then-apply path instead of writing directly to a target that Reprint cannot inspect after the process dies.

`pull-db` handles SIGINT and SIGTERM between SQL chunks, exits with code 2, and leaves the current stage unfinished. Running the command again continues it. Repeating an already completed `db-pull` or `db-apply` reports the saved result instead of throwing an error.

Reprint still stops when SQL was written to stdout, `db.sql` or the target settings changed, a required file is missing, or an actual read/write operation fails. In those cases continuing would risk skipping or repeating unknown work.

This is a single-PR alternative to the #590 → #591 → #595 → #582 → #607 stack.

## Testing

The focused PHP suites cover exit code 2, process death, repeated completed commands, saved file sizes, changed inputs, SQL replay, and wrapper signal handling. The MySQL tests cover a lingering old connection, SQL mode and foreign-key setup on the replacement connection, and the download-then-apply path.


adamziel:

PR #619, opened 2026-08-14
[PHP] Reapply MySQL dump settings after reconnect

adamziel:
A resumed MySQL import now reloads the dump's initial `SET` statements from a local file before continuing after the dump header.

## Problem

These settings belong to one MySQL connection. A dump starts with:

```sql
SET @OLD_UNIQUE_CHECKS=@@UNIQUE_CHECKS, UNIQUE_CHECKS=0;
SET @OLD_FOREIGN_KEY_CHECKS=@@FOREIGN_KEY_CHECKS, FOREIGN_KEY_CHECKS=0;
SET @OLD_SQL_MODE=@@SQL_MODE, SQL_MODE='ONLY_FULL_GROUP_BY,ERROR_FOR_DIVISION_BY_ZERO';
SET AUTOCOMMIT=0;
```

If `db-pull` or `db-apply` stops after that header, the next process opens a new MySQL connection. The new connection does not inherit those settings. Continuing from the saved position can then reject rows because of foreign keys or a different SQL mode.

## This change

The first SQL response includes the setup block as a separate part. The client saves it as `db-session-setup.sql` beside `db.sql`.

A direct-MySQL `db-pull` continuing from a saved source position runs that local file before requesting more SQL. A file-backed `db-apply` runs the same local file before seeking to its saved position. An import starting from the beginning still runs the normal header in `db.sql`.

This PR only restores the connection settings. #620 stores the source position alongside committed target data. #621 restarts an unfinished table from its `DROP TABLE` and `CREATE TABLE` statements.

## Testing

One end-to-end case kills a direct-MySQL `db-pull` after it saves a source position. Another downloads `db.sql`, kills `db-apply` after it saves a file position, and resumes from the file. Both continue on a new MySQL connection and import an ENUM value that the target's default SQL mode rejects, but the saved dump SQL mode accepts.

### DIFF
--- a/packages/reprint-server/src/class-mysql-dump-producer.php
+++ b/packages/reprint-server/src/class-mysql-dump-producer.php
@@ -400,13 +400,17 @@ class MySQLDumpProducer
      */
     private function emit_sql_header()
     {
-        $header =
-            "SET @OLD_UNIQUE_CHECKS=@@UNIQUE_CHECKS, UNIQUE_CHECKS=0;\n" .
+        $this->current_sql_fragment = self::get_session_setup_sql();
+    }
+
+    /** Returns the connection settings required before executing dump SQL. */
+    public static function get_session_setup_sql()
+    {
+        return "SET @OLD_UNIQUE_CHECKS=@@UNIQUE_CHECKS, UNIQUE_CHECKS=0;\n" .
             "SET @OLD_FOREIGN_KEY_CHECKS=@@FOREIGN_KEY_CHECKS, FOREIGN_KEY_CHECKS=0;\n" .
             // @TODO: Restore STRICT_TRANS_TABLES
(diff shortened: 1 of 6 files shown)


packages/reprint-server/src/class-mysql-dump-producer.php
```diff
@@ -400,13 +400,17 @@ private function emit_create_table_statement()
      */
     private function emit_sql_header()
     {
-        $header =
-            "SET @OLD_UNIQUE_CHECKS=@@UNIQUE_CHECKS, UNIQUE_CHECKS=0;\n" .
+        $this->current_sql_fragment = self::get_session_setup_sql();
+    }
+
+    /** Returns the connection settings required before executing dump SQL. */
+    public static function get_session_setup_sql()
```
adamziel, line 407: This work but is awkward, let's revisit this later.

PR #620, opened 2026-08-14
[PHP] Resume db-pull from the cursor saved in MySQL

adamziel:
A later `db-pull --sql-output=mysql` process now continues from the cursor committed in the target MySQL database instead of starting the dump again.

## Why the old local buffer was unsafe

The old importer kept an unfinished SQL statement in `pull/sql-buffer`. That file and the cursor were saved separately from the target database. After a crash, Reprint could append repeated SQL bytes, continue past InnoDB rows which MySQL rolled back, or repeat a MyISAM statement after MyISAM had already kept some rows.

This PR removes that file. An unfinished statement exists only in the current importer process. If that process stops, the next process requests the statement again from the last cursor committed in MySQL.

For InnoDB, imported rows and the cursor commit or roll back together.

For MyISAM, a stopped multi-row `INSERT` may keep a complete row prefix:

```sql
INSERT INTO posts (id, title) VALUES
    (1, 'one'),
    (2, 'two'),
    (3, 'three')
ON DUPLICATE KEY UPDATE id = id;
```

If MyISAM keeps rows 1 and 2 before the connection stops, the cursor still points before this statement. The next process requests the statement again. Rows 1 and 2 take the no-op duplicate-key path, and row 3 is inserted.

## How resume works

After each complete multipart SQL part, Reprint writes that part's cursor to `__reprint_db_pull_progress_<uuid>` and commits. A later process waits for the old target connection to finish, reads that cursor, reruns `db-session-setup.sql`, and asks the source for the following SQL.

The target lock prevents an older MySQL query from changing a table while the later process continues. A fresh import clears the target cursor before recording its SQL stage locally, so a stop during setup cannot revive an older import.

Reprint stops only when the next target operation cannot be repeated safely. This currently means a nontransactional table without a non-null unique key, or a nontransactional oversized value assembled through separate `UPDATE ... CONCAT()` statements. The error names the table and asks for `db-pull --abort` instead of guessing.

The progress-table UUID identifies its schema. Reprint logs and excludes that internal table name if the source database is itself a Reprint target.

This PR uses one completed producer action per multipart SQL part so the cursor rule is easy to test. The stacked #621 restores the exporter's existing fragment, time, and memory budgeting, allowing several regular `INSERT` statements to share one part while keeping table replacement and oversized-value updates separate.

## Testing

The MySQL end-to-end test kills the importer after a committed InnoDB part, holds the target lock while the next process starts, kills it again after a keyed MyISAM part, and then completes in a third process. It checks that neither later process sends the earlier table's `DROP TABLE`, that both tables contain every source row, and that the saved MySQL cursor matches the imported rows.

Stack: merged #619 and #639 → this PR → #621

### DIFF
--- a/docs/PUSH-TERMINOLOGY.md
+++ b/docs/PUSH-TERMINOLOGY.md
@@ -192,8 +192,7 @@ their parent directories.
         │   ├── remote-index.next.jsonl
         │   ├── fetch-list.jsonl
         │   ├── volatile-files.json
-        │   ├── sql-stats.json
-        │   └── sql-buffer
+        │   └── sql-stats.json
         └── push/
 ```
 
--- a/docs/PUSH-TERMINOLOGY.md
+++ b/docs/PUSH-TERMINOLOGY.md
@@ -211,7 +211,6 @@ Use these path names:
 | Fetch list file | `$fetch_list_file` |
 | Volatile files file | `$volatile_files_file` |
 | SQL statistics file | `$sql_stats_file` |
-| SQL buffer file | `$sql_buffer_file` |
 | Audit log file | `$audit_log_file` |
 | Progress file | `$progress_file` |
 | Reprint process lock path | `$process_lock_path` |
--- a/packages/reprint-client/src/import.php
+++ b/packages/reprint-client/src/import.php
@@ -7934,7 +7934,10 @@ class ImportClient
                     $source_table_name = is_array($source_table)
                         ? $source_table["name"] ?? null
                         : null;
-                    if ($source_table_name === $table) {
+                    if (
+                        is_string($source_table_name)
+                        && strcasecmp($source_table_name, $table) === 0
+                    ) {
                         throw new RuntimeException(
                             "The source database has a table named {$table}, which Reprint needs " .
                             "to save db-pull progress. Rename that source table before using " .
--- a/packages/reprint-client/src/import.php
+++ b/packages/reprint-client/src/import.php
@@ -176,6 +176,7 @@ class ImportClient
     private const STATE_PATH_ENCODING_PREFIX = "base64:";
     private const SQLITE_PREPARED_INSERT_CACHE_MAX = 128;
     private const MYSQL_DB_PULL_PROGRESS_TABLE = "__reprint_db_pull_progress";
+    private const MYSQL_DB_PULL_PROGRESS_TABLE_COMMENT = "Reprint db-pull progress v1";
 
     /**
      * Maximum number of consecutive interrupted responses with no cursor
@@ -7914,6 +7915,36 @@ class ImportClient
             $user = $this->mysql_user ?? "root";
             $pass = $this->mysql_password ?? "";
             $name = $this->mysql_database;
+            $table = self::MYSQL_DB_PULL_PROGRESS_TABLE;
+
+            $tables_file = wp_join_unix_paths($this->state_dir, "db-tables.jsonl");
+            $tables_handle = @fopen($tables_file, "r");
+            if (!$tables_handle) {
+                throw new RuntimeException(
+                    "Cannot check the source table names because db-tables.jsonl cannot be read."
+                );
+            }
+            try {
+                while (true) {
+                    $table_line = fgets($tables_handle);
+                    if ($table_line === false) {
+                        break;
+                    }
+                    $source_table = json_decode($table_line, true);
+                    $source_table_name = is_array($source_table)
+                        ? $source_table["name"] ?? null
+                        : null;
+                    if ($source_table_name === $table) {
+                        throw new RuntimeException(
+                            "The source database has a table named {$table}, which Reprint needs " .
+                            "to save db-pull progress. Rename that source table before using " .
+                            "--sql-output=mysql."
+                        );
+                    }
+                }
+            } finally {
+                fclose($tables_handle);
+            }
 
             // Parse host for port/socket (same format as WordPress DB_HOST).
             // An explicit --mysql-port takes precedence over a port embedded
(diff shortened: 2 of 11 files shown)


docs/PUSH-TERMINOLOGY.md
```diff
@@ -193,8 +193,7 @@ their parent directories.
         │   ├── fetch-list.jsonl
         │   ├── volatile-files.json
         │   ├── domains.json
-        │   ├── sql-stats.json
-        │   └── sql-buffer
+        │   └── sql-stats.json
```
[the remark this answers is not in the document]
adamziel: Addressed in 0447d8ce. Removed the stale SQL buffer file row from the terminology table.

packages/reprint-client/src/import.php
```diff
@@ -7632,44 +7633,30 @@ private function fetch_sql(): void
...
-                $this->audit_log(
-                    "SQL OUTPUT mysql | ran saved session setup after reconnect",
-                    true,
+            // phpcs:disable WordPress.Security.EscapeOutput.ExceptionNotEscaped -- MySQL errors are CLI text, not HTML.
+            $table = self::MYSQL_DB_PULL_PROGRESS_TABLE;
+            $create_table_sql = "CREATE TABLE IF NOT EXISTS `{$table}` (" .
+                "`id` TINYINT UNSIGNED NOT NULL PRIMARY KEY," .
+                "`source_hash` CHAR(64) CHARACTER SET ascii NOT NULL," .
+                "`source_cursor` MEDIUMTEXT NOT NULL" .
+                ") ENGINE=InnoDB";
```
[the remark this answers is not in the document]
adamziel: Addressed in 0447d8ce. The table now carries a Reprint marker and its engine and columns are checked before any row is deleted. An unrelated target table is left unchanged. db-pull also reads db-tables.jsonl first and rejects a source table with the reserved name before connecting to the target.

packages/reprint-client/src/import.php
```diff
@@ -7835,24 +7823,13 @@ private function fetch_sql(): void
...
 
                                 if ($query_complete) {
                                     $this->execute_mysql_queries($mysql_conn, $sql_buffer);
-
-                                    // Query executed — truncate the buffer file and reset.
-                                    if ($sql_buffer_handle) {
-                                        ftruncate($sql_buffer_handle, 0);
-                                        rewind($sql_buffer_handle);
-                                    }
+                                    $this->save_mysql_output_cursor($mysql_conn, $cursor);
```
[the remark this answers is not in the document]
adamziel: Addressed in 0447d8ce. The exporter now ends each multipart SQL part at the end of a statement, even when the tuner asks for 5,000 fragments. An INSERT can span responses, but it cannot share its completed part with the next DROP or CREATE. The E2E now requests 5,000 fragments instead of forcing one.

tests/e2e/tests/import-57-mysql-target-position.test.js
```diff
@@ -0,0 +1,221 @@
...
+            );
+            assert.ok(
+                Number(importedRows.rowCount) > 0 && Number(importedRows.rowCount) < 600,
+                `expected a partially imported table, got ${importedRows.rowCount} rows`,
+            );
+            const [[savedPosition]] = await interruptedTarget.query(
+                'SELECT source_cursor FROM `__reprint_db_pull_progress` WHERE id = 1'
+            );
+            assert.equal(typeof savedPosition.source_cursor, 'string');
+            assert.ok(savedPosition.source_cursor.length > 0);
```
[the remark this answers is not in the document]
adamziel: Addressed in 0447d8ce. The test decodes the saved producer cursor and checks its table, primary-key columns, and last id. The saved id must equal the row count visible from another MySQL connection.

packages/reprint-client/src/import.php
```diff
@@ -7506,6 +7508,36 @@ private function fetch_sql(): void
...
+                while (true) {
+                    $table_line = fgets($tables_handle);
+                    if ($table_line === false) {
+                        break;
+                    }
+                    $source_table = json_decode($table_line, true);
+                    $source_table_name = is_array($source_table)
+                        ? $source_table["name"] ?? null
+                        : null;
+                    if ($source_table_name === $table) {
```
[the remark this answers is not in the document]
adamziel: Addressed in a510d03a. The source-table check now uses a case-insensitive comparison, so names such as `__REPRINT_DB_PULL_PROGRESS` are rejected before the target connection opens. The E2E uses that uppercase variant.

packages/reprint-client/src/import.php
```diff
@@ -7506,6 +7508,39 @@ private function fetch_sql(): void
...
+                    }
+                    $source_table = json_decode($table_line, true);
+                    $source_table_name = is_array($source_table)
+                        ? $source_table["name"] ?? null
+                        : null;
+                    if (
+                        is_string($source_table_name)
+                        && strcasecmp($source_table_name, $table) === 0
+                    ) {
+                        throw new RuntimeException(
```
adamziel: Let's just log that for the user to see and skip that table. It's okay for that remote site to be itself imported through reprint, we don't need to stop the transfer.

packages/reprint-client/src/import.php
```diff
@@ -172,6 +172,8 @@ class ImportClient
     private const SAVE_STATE_EVERY_N_CHUNKS = 50;
     private const STATE_PATH_ENCODING_PREFIX = "base64:";
     private const SQLITE_PREPARED_INSERT_CACHE_MAX = 128;
+    private const MYSQL_DB_PULL_PROGRESS_TABLE = "__reprint_db_pull_progress";
```
adamziel: Let's revisit this later in a follow-up PR. [The URL rewrite command](https://github.com/WordPress/reprint/pull/578) also stores some of its progress in the database, we could use the same table for both.

packages/reprint-client/src/import.php
```diff
@@ -7527,44 +7562,104 @@ private function fetch_sql(): void
...
+            $table_result->free();
+            $table_engine = strtoupper($table_info["ENGINE"] ?? "");
+            $table_comment = $table_info["TABLE_COMMENT"] ?? "";
+            if (
+                !$table_info ||
+                $table_engine !== "INNODB" ||
+                $table_comment !== self::MYSQL_DB_PULL_PROGRESS_TABLE_COMMENT
+            ) {
+                throw new RuntimeException(
+                    "The target database already has a table named {$table}, and Reprint did not create it. " .
```
adamziel: This is oddly specific. I get that table comment mechanism but I'm not a fan of it. Let's just include some uuid in the table name and then we can be sure Reprint created it, or someone who wanted to imitate reprint. Let's eagerly drop and recreate that table on schema mismatch. We own it.

packages/reprint-client/src/import.php
```diff
@@ -7527,44 +7562,104 @@ private function fetch_sql(): void
...
+                throw new RuntimeException(
+                    "The target table {$table} does not have the columns Reprint expects. " .
+                    "Remove it before using --sql-output=mysql."
+                );
+            }
+
+            if (!$mysql_conn->query("DELETE FROM `{$table}` WHERE `id` = 1")) {
+                throw new RuntimeException(
+                    "MySQL could not reset the previous db-pull position: " . $mysql_conn->error
+                );
```
adamziel: This is too specific. Couldn't we just compare `SHOW CREATE TABLE`, if we want to compare anything at all? Also this could be a separate method such as `ensure_progress_table()` or so.

packages/reprint-client/src/import.php
adamziel: Addressed in bb27a68. db-pull now logs the source progress table and sends its name with every SQL request. The exporter skips that table case-insensitively, so a site that was already a Reprint target can still be imported.

packages/reprint-client/src/import.php
adamziel: Addressed in bb27a68. The table name now includes a UUID. The marker comment is gone, and Reprint drops and recreates the table when its schema does not match.

packages/reprint-client/src/import.php
adamziel: Addressed in bb27a68. The schema check now lives in ensure_progress_table(). It compares SHOW CREATE TABLE for the target table with a temporary table created from the expected definition.

packages/reprint-client/src/import.php
adamziel: Simplified further in 97ada15. The UUID now identifies the schema version. A schema change gets a new UUID and therefore a new table name, so there is no schema repair path.

packages/reprint-client/src/import.php
adamziel: Simplified further in 97ada15. Removed SHOW CREATE TABLE, the temporary comparison table, and ensure_progress_table(). CREATE TABLE IF NOT EXISTS now sits directly in the MySQL setup path.

PR #627, opened 2026-08-14
[PHP] Resume transactional table imports after committed SQL groups

adamziel:
Transactional tables now continue after their last committed SQL group instead of rebuilding the entire unfinished table.

#621 safely replays every unfinished table from `DROP TABLE` and `CREATE TABLE`. That is required for MyISAM, but it can repeat millions of already committed InnoDB rows after an importer process stops.

The exporter now includes the replaced table name with the replacement action. After creating the target table, the importer asks the target which storage engine it actually selected. Transactional engines clear the table-start position and continue from the source position committed with their rows. Nontransactional engines retain the table-start position and continue to rebuild the unfinished table.

For example:

```text
InnoDB: rows 1–20,000,000 committed → continue at row 20,000,001
MyISAM: process stops while filling table → run DROP + CREATE and refill that table
```

The InnoDB end-to-end case stops the importer after a committed batch and checks that the replacement process does not send a second `DROP TABLE`. #621's MyISAM case continues to require a complete table rebuild.

Stack: #619 → #620 → #621 → this PR

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -168,7 +168,7 @@ Entries in `paths_to_remove` under `wp-content/plugins/` also trigger automatic 
 
 ### SQL Streaming Crash Recovery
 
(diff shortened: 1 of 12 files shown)


adamziel:
Closing because #620 now resumes both InnoDB and keyed MyISAM tables from the cursor saved in MySQL. The table-rebuild split this PR depended on is no longer part of the design. #621 now handles batching through the exporter's existing fragment, time, and memory budgets.

PR #630, opened 2026-08-15
[PHP] Let files-pull mirror the current remote tree

adamziel:
`files-pull` now makes the local tree match the current remote tree by default. A local edit or deletion is restored when the remote path is unchanged, and a local-only path is removed. `--mode=catch-up` retains the previous delta-only behavior.

## This change

Mirror mode first applies the existing remote delta. It then reuses the local index mapper, patch planner, cleanup processor, and fetch pipeline to compare the last completed local index with a fresh local scan. Remote paths needed for replacement stay attached to their mapped local entries, so remapped files are fetched from the correct source.

The mode is saved with files-pull state and cannot change during resume. A verified local index is installed through the existing copy-and-swap path.

This first mirror mode is deliberately full-tree only. It rejects partial include, exclude, and filter selections, `--include-caches`, and `preserve-local`; those cases must use `--mode=catch-up`.

## Testing

Run an initial pull, then edit and delete remote-backed local files and add a local-only file. A second default pull should restore the remote-backed files and remove the local-only file. Repeat with `--mode=catch-up`; the three local changes should remain.

### DIFF
--- a/packages/reprint-client/src/import.php
+++ b/packages/reprint-client/src/import.php
@@ -88,6 +88,8 @@ require_once __DIR__ . '/lib/merge/load.php';
 require_once __DIR__ . '/lib/sort-index-file.php';
 require_once __DIR__ . '/lib/local-index-update-functions.php';
 require_once __DIR__ . '/lib/index/class-file-index-diff-processor.php';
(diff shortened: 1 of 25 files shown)


adamziel:

PR #637, opened 2026-08-15
Report live URL rewrite rows that require verification

adamziel:
`db-rewrite-urls` now writes a JSONL report for rows whose update result cannot be confirmed after the process dies.

## Background

MySQL may finish a row update after the PHP process has been killed, but before the command saves its next cursor. On resume, the same conditional update changes zero rows. That can mean the earlier update committed, or that another connection changed the row.

## This change

Each job creates `db-rewrite-urls-<job-id>.jsonl` beside its saved progress. The first line records the URL mappings and password-free database target. A zero-row conditional update appends the table, typed primary key, and hashes of the original and intended column values before the cursor advances. Full database values are not copied into the report.

A stable row ID prevents the final line from being appended twice. Resume also removes an incomplete final line before continuing.

Example completion:

```
db-rewrite-urls complete (5 records checked, 5 changed, 5 rows to verify: /path/to/db-rewrite-urls-<job-id>.jsonl)
```

## Testing

The E2E test sends `SIGKILL` to five consecutive CLI processes while MySQL is executing five different row updates, then resumes the job. It checks the complete JSONL report, every rewritten database value, untouched binary bytes, and `CHECK TABLE`.

The same five-crash sequence passes with InnoDB and MyISAM. Focused PHPUnit, PHP compatibility, PHPCS, PHPStan, and `git diff --check` also pass.

### DIFF
--- a/docs/DB-REWRITE-URLS.md
+++ b/docs/DB-REWRITE-URLS.md
@@ -66,3 +66,17 @@ https://example.com -> https://example.com/new
 The tradeoff is that a value written at the same time may still contain the old
 URL. If you need to catch every value, prevent the site from writing to the
 database while this command runs.
+
+## Rows to verify
+
+Every run creates a `db-rewrite-urls-<job-id>.jsonl` report in the same
+directory as its saved progress. The first line records the URL replacements
+and database used by the job.
+
+When the command cannot tell whether it updated a row before an interruption or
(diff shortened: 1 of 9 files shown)


adamziel:
This solves for a scenario that may or may not come up in practice at the expense of additional complexity. Let's keep an eye on it and revive this PR if needed.

PR #638, opened 2026-08-15
[PHP] Let file sync callers decide shared path copies

adamziel:
`PushPlan` and files-pull mirror need different answers for the same shared path. For example, `PushPlan` decides whether `uploads/a.jpg` changed from two local type, size, and ctime records. Mirror compares a fresh local entry with a mapped remote entry, whose ctime came from another machine, and uses a separate retained-to-fresh local plan to know whether the local path changed.

This PR keeps tree-shape decisions in `FileSyncPatchPlanner`: additions, deletions, collapsed subtree deletions, and type replacements still produce the same operations. For a path with the same shape in both indexes, the caller may now say whether that path needs copying. That includes an empty directory whose retained local-index row must be restored. Passing `null` keeps the existing type, size, and ctime comparison, so `PushPlan` behavior and its cursor remain unchanged.

The planner also accepts a separate decoder for each index and exposes the complete current entries, including decoder-added fields. A resumed planner must receive the same decoders; they are deliberately not added to the cursor. `PushPlan` exposes its existing active-deletion-roots work-file path so #633 can reuse that closed scratch file instead of creating another artifact.

This sits on #636, which provides the same independent-decoder and complete-entry API at the lower `FileIndexDiffProcessor` layer. The follow-up in #633 will replace its local-versus-mapped-remote classification loop with this planner.

## Testing

```sh
cd tests
../vendor/bin/phpunit Import/FileIndexDiffProcessorTest.php Import/FileSyncPatchPlannerTest.php Import/PushPlanTest.php
composer analyze -- --memory-limit=1G
git diff --check
```

### DIFF
--- a/packages/reprint-client/src/lib/index/class-file-sync-patch-planner.php
+++ b/packages/reprint-client/src/lib/index/class-file-sync-patch-planner.php
@@ -80,12 +80,14 @@ require_once __DIR__ . '/class-file-index-diff-processor.php';
  *     }
  *     $planner->close();
  *
- * The cursor contains everything resume() needs: both index paths, the active
- * deletion roots file, path selection, and the current byte offsets. The
- * active deletion roots file may contain bytes beyond the cursor after an
- * interruption. resume() ignores those bytes.
+ * The cursor contains both index paths, the active deletion roots file, path
+ * selection, and the current byte offsets. It does not store optional line
+ * decoders; pass the same decoders to resume(). The active deletion roots file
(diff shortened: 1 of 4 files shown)


adamziel:
Superseded by #633. This planner layer removed only 21 lines from the mirror caller. #633 now stacks directly on #636 and instead reuses the existing index sorter with an interruption-safe replacement fetch list, removing the separate candidate sort, anti-join, and append pipeline. Closing this PR keeps the stack to the two changes with clear callers.

PR #645, opened 2026-08-16
[PHP] Record local-only pull index deletions

adamziel:
The pull index journal can now record a deletion which changes the retained local index without changing the remote index.

## Example

Suppose `/workspace/debug.log` was created locally after the last pull. Mirror removes it because the current remote index has no `debug.log`.

Before this PR, the deletion API required both paths:

```php
$journal->record_successful_deletion(
    '/srv/site/debug.log',
    '/workspace/debug.log'
);
```

That call would claim a remote path was deleted and would remove it from the retained remote index. There is no correct remote path in this example because `debug.log` never existed remotely.

After this PR, mirror can record exactly what happened:

```php
$journal->record_local_deletion('/workspace/debug.log');
```

The WAL row contains only the local path:

```json
{"op":"-","local_relative_path_b64":"ZGVidWcubG9n"}
```

Local-index replay removes `debug.log`. Remote-index replay sees that no remote path is present and leaves the remote index unchanged.

## Testing

The focused test starts with separate local and remote index rows, records a local-only deletion, applies the WAL, and checks that the local index changes while the remote index stays byte-for-byte unchanged.

### DIFF
--- a/packages/reprint-client/src/lib/pull/class-pull-index-journal.php
+++ b/packages/reprint-client/src/lib/pull/class-pull-index-journal.php
@@ -362,6 +362,26 @@ class PullIndexJournal
     }
 
     /**
+     * Adds a `-` record after mirror removes a local-only path.
+     *
+     * @param string $local_absolute_path Local path already removed.
+     * @throws RuntimeException When the record cannot be appended.
+     */
+    public function record_local_deletion(string $local_absolute_path): void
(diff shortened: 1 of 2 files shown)


adamziel:
Folded this journal change into #633 so record_local_deletion() lands with its first production caller. #646 now targets trunk directly.

PR #646, opened 2026-08-16
[PHP] Remove pull paths without following parent symlinks

adamziel:
Remote pull deletions now remove a parent symlink instead of traversing it.

## Example

Assume this local tree:

```text
/workspace/link -> /outside
/outside/file.txt
```

The remote index says `/link/file.txt` was deleted.

Before this PR, recursive removal inspected `/workspace/link/file.txt`. The parent component was a symlink, so that path resolved to `/outside/file.txt`. The outside file could be removed while `/workspace/link` remained.

After this PR, the pull checks each component between `/workspace` and the requested path:

```text
/workspace/link/file.txt
           ^ first symlink component
```

It removes `/workspace/link` and stops. `/outside/file.txt` remains unchanged.

The call site now uses one result for all three safe outcomes:

```php
$removal = $this->remove_local_path_without_traversing_symlinks(
    $local_absolute_path
);
```

The result says whether the path was already absent, the requested path was removed, or a parent symlink was removed. PR #633 uses the same path for locally changed files which are missing from the current remote index.

## Testing

The focused test creates a parent symlink to an outside directory, applies a remote deletion below it, and checks that the symlink is gone while the outside file remains.

### DIFF
--- a/packages/reprint-client/src/import.php
+++ b/packages/reprint-client/src/import.php
@@ -8088,11 +8088,32 @@ class ImportClient
     }
 
     /**
-     * Removes a remote deletion root from the mapped local filesystem.
+     * Removes a local path without traversing a symlink component.
      *
-     * @return string|null The mapped local absolute path when it is absent
-     *                     after this call, or null when it could not be removed.
+     * @return array{path:string,outcome:'absent'|'removed'|'symlink'}|null
      */
+    private function remove_local_path_without_traversing_symlinks(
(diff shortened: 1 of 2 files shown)


adamziel:
Closing this PR and removing it from the mirror stack. The parent-symlink case can also hide unchanged sibling paths, so its behavior needs a separate decision in #650. PR #633 now targets trunk directly and does not try to handle paths through parent symlinks.

PR #661, opened 2026-08-17
[PHP] Share database import target preparation

adamziel:
Makes live `db-pull --sql-output=mysql` and file-based `db-apply` use one function to prepare a new or resumed database target.

Before this PR, both readers separately performed the same target work:

```text
fetch_sql()                     apply_database_dump_file()
    clear or read cursor            clear or read cursor
    check whether SQL can repeat    check whether SQL can repeat
    reload MySQL SET statements     reload MySQL SET statements
```

The readers still handle their different inputs. `fetch_sql()` reads multipart SQL from HTTP. `apply_database_dump_file()` reads marked SQL groups from `db.sql`. Both now call `prepare_database_import_target()` before sending groups through the existing shared importer:

```text
HTTP reader ───┐
               ├─> prepare target ─> execute SQL group and save its next cursor
db.sql reader ─┘
```

This keeps cursor reset and lookup, the MyISAM repeat check, and the saved MySQL session setup in one place. It does not change SQL grouping, transactions, cursor storage, or the two input readers.

## Testing

The focused SQLite import and database adapter unit tests pass. The existing MySQL session restart E2E covers both callers: direct MySQL `db-pull` and `db-apply` from `db.sql`.

### DIFF
--- a/packages/reprint-client/src/import.php
+++ b/packages/reprint-client/src/import.php
@@ -6385,10 +6385,6 @@ class ImportClient
     public function run_db_apply(array $options): void
     {
         $sql_file = wp_join_unix_paths($this->state_dir, "db.sql");
-        $session_setup_file = wp_join_unix_paths(
-            $this->state_dir,
-            "db-session-setup.sql",
-        );
         if (!file_exists($sql_file)) {
             throw new RuntimeException(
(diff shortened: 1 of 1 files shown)


adamziel:
Closing for now. The remaining duplication is small and does not justify a separate refactor today; the branch can remain available if we revisit it later.

PR #666, opened 2026-08-18
Remove dependencies for server component and move to `WordPress\Reprint\Server` namespace

fredrikekelund:
https://github.com/Automattic/jetpack/pull/50369 shipped the Reprint exporter in Jetpack, but with notable flaws. We needed two follow-up PRs and ultimately a revert when we discovered that Jetpack's security scanner flagged `plugins/jetpack/vendor/wp-php-toolkit/data-liberation/vendor-patched/symfony/polyfill-intl-normalizer/Normalizer.php` as potentially malicious. For the record, that was a false positive, but still clearly something we needed to address.

Moreover, as pointed out in https://github.com/Automattic/jetpack/issues/51027, the original Jetpack PR included a bunch of dependency code outside the pure Reprint server implementation. This code should ideally only be required when a Reprint client calls the appropriate Jetpack API endpoints, but the Jetpack autoloader is the joker in the mix that makes this much less pure. The GitHub issue documents at least one known case where the added dependencies led to name collisions.

The end goal of this PR is to make it safe to reinstate an updated version of https://github.com/Automattic/jetpack/pull/50369. To achieve this, I have done the following:

1. The _only_ function that the Reprint server was using from its Composer dependencies was `wp_join_unix_paths()`, which comes from `wp-php-toolkit/filesystem`. `wp-php-toolkit/reprint-server` did not directly require this package, so there was an inherent flaw there. My solution to this was to inline `wp_join_unix_paths()` and remove both Composer dependencies (`wp-php-toolkit/data-liberation` and `wp-php-toolkit/html`).
2. The Reprint server code was squatting `WordPress\DataLiberation` and the global namespace. This is an inherent risk. I refactored the code to consistently use `WordPress\Reprint\Server`. The fact that we recently renamed the package from `wp-php-toolkit/reprint-exporter` to `wp-php-toolkit/reprint-server` actually comes in handy here, because wpcomsh currently deploys the old package with the old namespace. By changing the namespace, we guarantee that there'll be no naming collisions during the interim period when we ship the Reprint exporter in both wpcomsh and Jetpack.

If and when we merge this, we should bump the package version to 0.10.0. I learned today that Composer (and npm semver, for that matter) treat caret (`^`) version ranges differently below 1.0.0. So anyone requesting `wp-php-toolkit/reprint-server@^0.9.0` will not get 0.10.0. This works in our favor, as the namespace change breaks the API contract.


packages/reprint-server/src/utils.php
```diff
@@ -528,7 +573,69 @@ function assert_valid_path(string $path, string $label = "path"): void
...
 }
+}
 
-} // !function_exists guard
+// ---------------------------------------------------------------------------
+// Vendored from wp-php-toolkit/filesystem.
+//
+// This is a copy of WordPress\Filesystem\wp_join_unix_paths(), kept in sync by
+// hand. Do not "fix" it by importing the original: reprint-server must require
+// nothing but PHP.
```
adamziel, line 583: Very nice, I love this.

PR #690, opened 2026-08-21
[PHP] Preserve multibyte text across oversized MySQL chunks

adamziel:
Oversized MySQL character columns now finish without skipping or repeatedly requesting the end of a multibyte value.

## Background

MySQL `SUBSTRING()` counts characters for text columns, while the exporter tracked progress with `strlen()`, which counts bytes. For example, two `utf8mb4` emoji are two characters but eight bytes. After exporting those emoji, the old code saved byte offset 8 and asked MySQL to continue at character 9. MySQL returned an empty string, the byte offset stayed at 8, and the exporter repeated that request until it timed out.

## This change

Character columns now keep separate byte and character offsets. The character offset selects the next MySQL substring. The returned chunk's `CHAR_LENGTH()` advances that offset, while its byte length advances the completion cursor.

The column collation supplies the maximum bytes per character. This bounds the requested character count so the encoded SQL stays within its byte limit. `VARCHAR`, `TEXT`, `MEDIUMTEXT`, and `LONGTEXT` use this path. `VARBINARY`, `BLOB`, `MEDIUMBLOB`, and `LONGBLOB` continue to use byte ranges by casting the value to binary before calling `SUBSTRING()`.

A saved cursor with partial character data from the old format cannot be converted safely, so it now asks the user to abort and restart the pull. An unexpected empty chunk before the saved byte length also stops with a clear error instead of looping. Only character and binary strings may use partial `CONCAT()` updates; an oversized `GEOMETRY` value is rejected before invalid SQL reaches the target database.

## Testing

The first commit contains only the regression tests. Against its parent implementation, the focused run reports eight failures and one error. The four binary controls pass.

The completed implementation passes 35 focused tests with 246 assertions. These cover all four character types and all four binary types, resume after every fragment, old cursors, a source value shrinking during export, mixed text and binary columns, statement byte limits, and oversized geometry rejection.

The full GitHub CI matrix passes, including PHP 5.6 through PHP 8.5, the PDO-less `wpdb` fallback, the pull pipeline benchmark, and Playground CLI E2E tests.

### DIFF
--- a/packages/reprint-server/src/class-database-rows-reader.php
+++ b/packages/reprint-server/src/class-database-rows-reader.php
@@ -57,12 +57,15 @@ class DatabaseRowsReader {
 
     /**
      * Column metadata cached by table and column name. Each column contains
-     * data_type (for example, varchar) and column_type (for example,
-     * varchar(255)).
+     * data_type (for example, varchar), column_type (for example,
+     * varchar(255)), and its nullable collation name.
      *
-     * @var array<string,array<string,array<string,string>>>
+     * @var array<string,array<string,array{data_type:string,column_type:string,collation:?string}>>
      */
     private $column_type_cache = [];
+
+    /** @var array<string,int> Maximum character bytes cached by collation. */
+    private $maximum_character_bytes_by_collation = [];
 
 
     /** @var array|null */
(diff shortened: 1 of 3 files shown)


adamziel:
@WPprodigy I don't see any way of safely resuming a migration with a cursor pointing mid-character. Updating the reprint client to this version will require restarting any in-progress migrations.

PR #691, opened 2026-08-21
Safeguard against commonly-disabled functions

WPprodigy:
Observed in the wild:

> The remote server crashed: Error: Call to undefined function disk_free_space()

Since PHP 8, a function listed in `disable_functions` is removed from the function table entirely, and calling it throws a fatal error. On PHP 7 the same call would only emit a warning and return null, which the `@` alone would have handled fine.

This PR adds some guarding for functions that are known to be disabled sometimes:

- `disk_free_space()` & `disk_total_space()`
- `ini_set()`
- Some rare-but-easy-to-protect ones: `php_sapi_name()` / `sys_get_temp_dir()`

### DIFF
--- a/packages/reprint-client/src/import.php
+++ b/packages/reprint-client/src/import.php
@@ -2942,6 +2942,23 @@ class ImportClient
             $all_pass = false;
         }
 
+        // 6. Response streaming won't be double-compressed
+        $limits = $data["limits"] ?? null;
+        $stream_ok = true;
+        if (is_array($limits) && isset($limits["zlib_output_compression_can_be_disabled"])) {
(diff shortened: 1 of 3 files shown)


WPprodigy:
Hmm, I think we need an preflight assertion check for this. If we can't turn off `zlib.output_compression` for example (and it's on), we should probably fail the preflight rather than continue.

adamziel:
@WPprodigy yes, disabling `zlib.output_compression` is not optional at the moment. I like your idea to fail early during the preflight phase. That should solve the problem at hand. Once we start seeing migrations failing due to host-enforced `zlib.output_compression=On`, we could explore an optional support for `zlib.output_compression=Off` that moves the data using more requests with a lower response size limit, e.g. 16MB or 32MB.

PR #692, opened 2026-08-22
Reprint Server: Prefer canonical Reprint API URLs in the client

atanas-dev:
Canonicalize the server API URL to `reprint-api`, ensuring the previous `site-export-api` is still supported for backwards-compatibility.


adamziel:
Thank you @atanas-dev !

PR #695, opened 2026-08-22
Reprint Server: Namespace the WordPress plugin runtime API

atanas-dev:
Namespace the WordPress plugin runtime and make sure the terminology used is up to date.

### DIFF
--- a/reprint-server-wp/compat.php
+++ b/reprint-server-wp/compat.php
@@ -1,0 +1,139 @@
+<?php
+
+/**
+ * Installs support for integrations which still use Site Export names.
+ *
+ * @param bool $normalize_request Whether to map the legacy query key for plugin routing.
+ */
+function reprint_server_bootstrap_compatibility(bool $normalize_request = true): void {
+    $constant_aliases = [
+        'SITE_EXPORT_VERSION' => 'WordPress\\Reprint\\Server\\Plugin\\VERSION',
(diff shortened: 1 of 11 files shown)


reprint-server-wp/compat.php
```diff
@@ -0,0 +1,139 @@
+<?php
+
+/**
+ * Installs support for integrations which still use Site Export names.
+ *
+ * @param bool $normalize_request Whether to map the legacy query key for plugin routing.
+ */
+function reprint_server_bootstrap_compatibility(bool $normalize_request = true): void {
```
adamziel: Do we need this? These things accumulate over time, it would be highly practical from Reprint's perspective to drop this and break BC on the next release. Would that create any deal-breaking inconveniences on the Jetpack side? cc @atanas-dev @WPprodigy

adamziel:
I left a note, I don't see any blockers here other than that 👍

reprint-server-wp/compat.php
```diff
@@ -0,0 +1,197 @@
+<?php
```
adamziel, line 1: Wait, did this file get longer than before?

adamziel:
I'm not a big fan of the compat.php file and how we keep the old names without keeping them, but let's move on with this. Maybe it will make the migration a bit easier for the updaters. Once they do, I'd like to still get rid of most of that.

PR #696, opened 2026-08-22
Reprint Server: Canonical plugin configuration with legacy migration.

atanas-dev:
Namespace and match the latest terminology in the plugin configuration, making sure old values are still read for backwards compatibility.

### DIFF
--- a/reprint-server-wp/compat.php
+++ b/reprint-server-wp/compat.php
@@ -67,6 +67,25 @@ function reprint_server_compat_expose_legacy_names(): void {
         },
         -PHP_INT_MAX
     );
+
+    // TODO: This filter should be deleted after September 2026, as it should no longer be relevant by then.
+    add_filter(
+        'reprint_server_managed_push_enabled',
+        static function ($enabled) {
+            if ($enabled !== null) {
+                return $enabled;
+            }
+            if (defined('SITE_EXPORT_PUSH_ENABLED')) {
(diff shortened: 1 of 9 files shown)


reprint-server-wp/lib.php
```diff
@@ -30,14 +30,14 @@
 if (!defined(__NAMESPACE__ . '\\PLUGIN_DIR')) {
     define(__NAMESPACE__ . '\\PLUGIN_DIR', plugin_dir_path(__FILE__));
 }
-if (!defined(__NAMESPACE__ . '\\SECRET_FILE')) {
-    define(__NAMESPACE__ . '\\SECRET_FILE', PLUGIN_DIR . 'secret.php');
+if (!defined(__NAMESPACE__ . '\\CONNECTION_TOKEN_FILE')) {
```
adamziel, line 33: A changelog entry would help the consumers of this code migrate. CC @WPprodigy @fredrikekelund for visibility for the next time this code is updated in Jetpack.

reprint-server-wp/compat.php
```diff
@@ -61,10 +61,155 @@ static function ($options) {
         -PHP_INT_MAX
     );
 
+    add_filter(
```
adamziel, line 72: I have the same BC note here as in the previous PR in this stack. If we can avoid accumulating complexity, that's brilliant. If that will be too much of an inconvenience in Jetpack/wpcomsh/another consumer, then let's keep it. Any thoughts @fredrikekelund?

PR #713, opened 2026-08-28
Remove `autoload.files` altogether from Reprint server

fredrikekelund:
As discussed in https://github.com/Automattic/jetpack/pull/51430#discussion_r3864891571, always autoloading `utils.php` creates unnecessary overhead and risk at Jetpack scale. Although the file contents are relatively harmless, it's better to be safe here.

This PR takes the cheaper approach and adds `require_once` statements for `utils.php` to the relevant files. We can do better and add a `Utils` class that's autoloaded via the classmap, but I chose the easy approach for this initial PR.

### DIFF
--- a/packages/reprint-client/src/import.php
+++ b/packages/reprint-client/src/import.php
@@ -66,16 +66,7 @@ foreach ([
 ] as $autoloader) {
     if (file_exists($autoloader)) {
         require_once $autoloader;
-        break;
-    }
-}
-
-foreach ([
-    __DIR__ . '/../../reprint-server/src/utils.php',
-    __DIR__ . '/../../../vendor/wp-php-toolkit/reprint-server/src/utils.php',
-] as $utils_path) {
-    if (file_exists($utils_path)) {
-        require_once $utils_path;
+        require_once dirname($autoloader) . '/wp-php-toolkit/reprint-server/src/utils.php';
         break;
     }
 }
--- a/packages/reprint-client/src/import.php
+++ b/packages/reprint-client/src/import.php
@@ -70,6 +70,8 @@ foreach ([
     }
 }
 
+require_once __DIR__ . '/../../reprint-server/src/utils.php';
+
 // Load vendored MySQL query stream (from sqlite-database-integration PR #264)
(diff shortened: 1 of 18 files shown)


reprint-server-wp/lib.php
```diff
@@ -523,7 +540,7 @@ function _site_export_handle_api_request(array $options = []): void {
                 );
             }
             $docroot = $canonical_docroot === '/' ? '/' : rtrim($canonical_docroot, '/\\');
-            $lexical_docroot = \WordPress\Reprint\Server\normalize_path(str_replace('\\', '/', $configured_docroot));
+            $lexical_docroot = _site_export_normalize_path(str_replace('\\', '/', $configured_docroot));
```
fredrikekelund, line 543: I thought this coupling was unnecessary. Better to accept some duplication between `_site_export_normalize_path` and `\WordPress\Reprint\Server\normalize_path`

packages/reprint-client/src/import.php
```diff
@@ -70,6 +70,16 @@
...
 
+foreach ([
+    __DIR__ . '/../../reprint-server/src/utils.php',
+    __DIR__ . '/../../../vendor/wp-php-toolkit/reprint-server/src/utils.php',
+] as $utils_path) {
+    if (file_exists($utils_path)) {
+        require_once $utils_path;
+        break;
+    }
+}
```
WPprodigy: Both of these need to be loaded?

fredrikekelund: The layout can legitimately differ depending on whether we are currently inside the built client PHAR (in which case the client code lives in `packages/reprint-client` and the server code lives in `vendor/wp-php-toolkit/reprint-server`) or not (in which case `wp-php-toolkit/reprint-client` and `wp-php-toolkit/reprint-server` are siblings inside a `vendor` directory).

I did revisit this by folding it into the existing `foreach` that looks up the `autoload.php` file, which makes it a lot neater.

tests/PushCommitTest.php
fredrikekelund, line 1: These changes are Codex resolving a race condition that was unrelated to my other changes in this PR

PR #714, opened 2026-08-28
Use the adaptive tuner to handle http 413 errors

WPprodigy:
A `files-pull` can fail with an HTTP 413 (request body too large), without us ever trying to scale it down.

With this PR, a `413` now diagnoses as `REQUEST_TOO_LARGE`, halves the tuner's request-body budget (800 KiB → 400 KiB → 256 KiB, taking the floor on the last attempt before the existing three-failure limit), and rebuilds the batch from
the same fetch-list offset so nothing is skipped.

The budget has a maximum of 80% of `max_request_bytes` (read from preflight) and is capped at 800 KiB regardless of what the host reports (we can see what PHP says, but not higher levels like nginx). This helps keeps things under the common 1mb nginx limit  with extra room for the multipart envelope. It costs basically no throughput: batch size doesn't set the pace, the number of requests needed to stream the files does and the batch is re-uploaded on every one of them.

### DIFF
--- a/packages/reprint-client/src/import.php
+++ b/packages/reprint-client/src/import.php
@@ -222,6 +222,7 @@ class ImportClient
     private const POTENTIALLY_TRANSIENT_HTTP_STATUS_CODES = [
         400, // Bad Request
         408, // Request Timeout
+        413, // Content Too Large (adaptive tuner will adjust)
         418, // Observed when an upstream bot filter replaced a Reprint response
         425, // Too Early
         429, // Too Many Requests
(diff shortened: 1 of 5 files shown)


adamziel:
Oh, we've ran into that on an actual site? This is so weird! How is it accepting media uploads? :o Scaling down and retrying sounds like a good idea, and at the same time making 800KiB a hard cap will prevent Reprint from sending larger  requests to hosts that support them—which matters a lot for push. Let's explore how we can support both large caps and small caps.

Edit: Push has a separate tuner so we're good in that regard. And maybe 800KiB as a hard limit for large lists of pull files isn't that bad after all—we likely won't download all their bodies in a single request anyway due to other limits. OK, I think we're good here! Thank you!

PR #718, opened 2026-08-31
[PHP] Normalize MariaDB zero-byte spatial values to NULL

adamziel:
Database pulls now preserve MariaDB `NOT NULL` spatial schemas on MariaDB targets. MySQL 8 targets receive SQL `NULL` instead.

## Background

MariaDB permits this change on a populated table:

```sql
ALTER TABLE maps ADD COLUMN location POINT NOT NULL;
```

Existing rows receive a zero-byte spatial value. Reprint emitted that value as `''`, which neither MariaDB nor MySQL accepts in an `INSERT`. The database apply then stopped with error 1416.

## This change

When the producer reads a zero-byte spatial value, it emits a marked SQL `NULL` expression and the marked fallback `ALTER TABLE`.

The dump does not change the source table. It keeps the original `CREATE TABLE`, writes any earlier ordinary `INSERT` statements, then carries the marked `ALTER TABLE` before the first zero-byte value for that column. A table can therefore appear as `CREATE → INSERT → ALTER → INSERT`; when the first row has the placeholder, the `ALTER` comes immediately after `CREATE`.

On a MariaDB target, the importer omits the marked column from its one-row `INSERT` and changes the marked `ALTER TABLE` into a no-op. With no explicit column default, MariaDB then writes its own zero-byte placeholder and the source `NOT NULL` definition stays unchanged.

On a MySQL target, the importer rebuilds the marked `ALTER TABLE` from the target's own `SHOW CREATE TABLE` result. It forces quoted `SHOW CREATE` output and removes only `NOT` from the column's top-level `NOT NULL`. The rest of the target definition remains intact, including comments, defaults, invisible-column state, and engine-specific attributes. The value is SQL `NULL`. An older importer can still execute the valid fallback `ALTER TABLE` carried in the dump.

The saved pull position records pending and completed column changes, so an interrupted pull can continue on either side of the `ALTER TABLE`.

## Limitations

This change recognizes only MariaDB's exact zero-byte placeholder. The following issues must still be resolved in stacked PRs:

* **Schema match:** MariaDB source → MariaDB target preserves the source `NOT NULL` schema. If the affected column has no explicit default, it also preserves its zero-byte value. MariaDB source → MySQL 8 target must use `NULL` in a nullable column because MySQL cannot represent the source placeholder; that path is not schema-exact.

* A MariaDB column with an explicit spatial default uses that default when Reprint omits the zero-byte value. The target schema matches, but that row does not retain the zero-byte value.
* On a MySQL target, a `CHECK` constraint which rejects SQL `NULL` cannot represent the normalized row. The import stops and leaves the constraint in place.
* On a MySQL target, a `SPATIAL INDEX` requires its column to remain `NOT NULL`. If a changed target has such an index, Reprint stops before altering the column; it cannot preserve the index and store `NULL` in the same column.
* Do not change a marked MySQL target column after the dump creates its table and before the nullable `ALTER TABLE` runs. The importer deliberately uses the target's current column definition so target-specific attributes survive.
* A spatial value larger than one SQL statement enters Reprint's general oversized-value path. That path appends fragments with `UPDATE ... CONCAT()`, but a spatial column cannot temporarily hold an incomplete geometry. The import therefore stops.
* Reprint preserves a valid geometry's exact bytes and SRID. It does not translate coordinate axis rules between engines. In a real MariaDB to MySQL 8 probe, `POINT(7 8)` with SRID 4326 reached MySQL with the same bytes and SRID. MySQL's default `ST_AsText()` returned `POINT(8 7)` because MySQL applies its SRID 4326 axis order. `ST_AsText(value, 'axis-order=long-lat')` returned `POINT(7 8)`.
* MariaDB can store an SRID integer which is absent from MySQL's spatial reference system registry. MySQL rejects that value during import.
* Any malformed nonzero spatial value is emitted unchanged. The target may reject it; Reprint does not silently convert it to `NULL`.

## Design notes

* Query rewriting is done in the importer to avoid shipping the SQL parser with the exporter plugin. It would be a heavy dependency.

## Testing

The tests create the zero-byte state on MariaDB and apply the same dump to MariaDB and Oracle MySQL 8. They cover all eight spatial types, SQL `NULL`, a real empty `GEOMETRYCOLLECTION`, a non-empty geometry collection, a self-intersecting polygon, SRIDs 0 and 4326, exact preservation of every nonzero spatial value, batching, and interrupted and resumed pulls. The fixture has no spatial defaults, so its MariaDB target keeps all eight `NOT NULL` columns and their zero-byte values. The MySQL target makes those columns nullable and uses `NULL`.

They also prove that MariaDB 10.11 rejects a spatial index over the zero-byte source value, that a MySQL target forces quoted `SHOW CREATE TABLE` output before parsing it, and that the importer names a MySQL spatial index which prevents the nullable rewrite.

The focused real-engine suite and both cross-engine E2E cases run in CI.


packages/reprint-server/src/class-mysql-dump-producer.php
```diff
@@ -369,6 +410,67 @@ private function emit_row()
...
+        foreach ($this->pending_nullable_spatial_columns as $column) {
+            $column_metadata = $this->row_reader->get_column_metadata($column);
+            $definition = "MODIFY COLUMN " . $this->row_reader->quote_identifier($column) .
+                " " . $column_metadata["column_type"] . " NULL";
+            if ($column_metadata["comment"] !== "") {
+                $comment = str_replace(
+                    ["\\", "'"],
+                    ["\\\\", "''"],
+                    $column_metadata["comment"]
+                );
```
adamziel: this is sus, shouldn't we quote this using a specialized method?

packages/reprint-server/src/class-mysql-dump-producer.php
```diff
@@ -369,6 +410,67 @@ private function emit_row()
...
+                $definition .= " COMMENT '{$comment}'";
+            }
+            $definitions[] = $definition;
+        }
+
+        $quoted_table = $this->row_reader->quote_identifier(
+            $this->row_reader->get_current_table()
+        );
+        $this->current_sql_fragment = "ALTER TABLE {$quoted_table}\n" .
```
adamziel: Isn't there a way to safely rewrite SQL using the SQL processor machinery?

packages/reprint-server/src/class-mysql-dump-producer.php
adamziel: Yes. Replaced the manual escaping with `$this->db->quote()` in 841daed9. The real-engine test now uses a comment containing an apostrophe, backslashes, and a newline and passes on both MariaDB and Oracle MySQL 8 targets.

packages/reprint-server/src/class-mysql-dump-producer.php
adamziel: Checked. This is generated DDL, not an existing statement being rewritten. The exporter package does not ship the MySQL parser. That parser yields an AST but has no MySQL mutation and serialization API, and the PHP 5.6 exporter build cannot install its PHP 7.2 runtime. Pulling it in for this fixed statement would add a large dependency without removing the need to build the ALTER. The dynamic pieces are constrained: identifiers use `quote_identifier()`, comments use the connection `quote()`, and `column_type` comes from `SHOW FULL COLUMNS` only after `data_type` passes the exact spatial allowlist.

packages/reprint-server/src/class-mysql-dump-producer.php
adamziel: Follow-up: the quoted comment remains only in the valid fallback for importers which do not recognize the marker. The current importer reuses the target full column definition, so it does not rebuild or requote the comment.

packages/reprint-server/src/class-mysql-dump-producer.php
adamziel: You were right. a442a0f9 moves the complete rewrite to the importer. The exporter adds a marker with base64 table and column names plus valid fallback SQL. The importer uses `WP_MySQL_Lexer` on the target `SHOW CREATE TABLE` result and removes only the top-level `NOT` token. The exporter needs no SQL parser.

packages/reprint-server/src/class-mysql-dump-producer.php
```diff
@@ -369,6 +410,67 @@ private function emit_row()
...
+            throw new \RuntimeException(
+                "Cannot emit a nullable spatial ALTER TABLE without any pending columns."
+            );
+        }
+
+        $definitions = [];
+        foreach ($this->pending_nullable_spatial_columns as $column) {
+            $column_metadata = $this->row_reader->get_column_metadata($column);
+            $definition = "MODIFY COLUMN " . $this->row_reader->quote_identifier($column) .
+                " " . $column_metadata["column_type"] . " NULL";
```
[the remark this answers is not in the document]
adamziel: Addressed in a442a0f9. The importer now builds `MODIFY COLUMN` from the target complete `SHOW CREATE TABLE` definition and removes only the top-level `NOT` token. The real MariaDB test checks that type, default expression, invisible state, and comment remain unchanged while nullability changes to `YES`.

PR #754, opened 2026-09-04
Add standalone source-host file cleanup command

adamziel:
Source-host mu-plugins can crash a destination before an integration's target runtime setup runs. Add a low-level `remove-host-files` command for integrations that manage their own target runtime.

The command runs after `preflight` and `files-pull`, reuses the detected host analyzer manifest, removes only declared host files, and records the removed paths. It does not generate runtime configuration or change the database. The existing `apply-runtime` path now shares the same removal logic.

Usage:

```bash
php reprint.phar remove-host-files "$URL" \
  --state-dir="$STATE_DIR" \
  --fs-root="$FS_ROOT"
```

Use `--flat-document-root="$DOCUMENT_ROOT"` instead of `--fs-root` when the pulled files have already been flattened into the document root.

Testing:
- `../vendor/bin/phpunit Import/ProductionDropInRemovalTest.php Import/WpengineHostAnalyzerTest.php Import/CliHelpTest.php` — 40 tests, 188 assertions
- `php -l packages/reprint-client/src/import.php`
- `git diff --check`

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -155,14 +155,14 @@ PHPUnit tests automatically create/drop test databases. The naming convention is
 
 ### Runtime Manifest and Host Analyzers
 
-The `apply-runtime` command separates source host detection from target runtime configuration. The flow is:
+The `apply-runtime` command separates source host detection from target runtime configuration. The `remove-host-files` command uses the same host-analyzer manifest when an integration provides its own target runtime. The flow is:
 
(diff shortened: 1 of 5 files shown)


adamziel:
Closing this without merging. We will handle the problem earlier by not downloading these source-host-specific MU-plugin files during the file pull, instead of downloading them and deleting them afterward. That avoids putting files on the destination that only work on the source host and avoids adding a separate cleanup command.

PR #763, opened 2026-09-05
[PHP] Create a one-site target network without renumbering the site

adamziel:
Lets a selected network site boot at a new URL in an empty MySQL database, keeping its site ID and table names.

This layer creates a **one-site network**. #770 changes that target to ordinary single-site WordPress and replaces this PR's new `--network-admin` option with `--site-admin`.

## Example at this layer

Create an empty MySQL database named `shop_local`. Replace `alice` with a login included in the selected site's export:

```sh
php reprint.phar pull 'https://network.example/shop/?reprint-api' \
  --secret="$NETWORK_TOKEN" --state-dir=./shop-state --fs-root=./shop-files \
  --target-engine=mysql --target-host=127.0.0.1 \
  --target-user="$MYSQL_USER" --target-pass="$MYSQL_PASSWORD" --target-db=shop_local \
  --new-site-url=http://localhost:9000 --network-admin=alice \
  --flatten-to=./shop --runtime=php-builtin --output-dir=./shop-runtime \
  --start-runtime=none

bash ./shop-runtime/start.sh
```

`--network-admin=alice` grants that imported account access to the new network. The target gets a new `wp-config.php` with target database credentials and fresh login salts. Existing uploads keep their paths, and sibling links keep pointing to the source network.

An existing target table stops the import before any imported `DROP TABLE` runs. Direct `db-pull --sql-output=mysql` is rejected because it skips target setup. After an interruption, rerun the same pull command without changing the target database, URL, or administrator.

## Testing

Open the target and an uploaded file. Try a second migration into a database containing a table; that table must remain intact. E2E tests also stop and resume the importer around database initialization, SQL import, and cleanup.

Based on #762.


adamziel:
Replaced by #770. Its combined change now targets trunk and creates ordinary single-site WordPress directly, while keeping the empty-target checks, URL and upload mapping, and resumable target setup from this PR. The remaining stack is #770 → #774 → #775 → #778.

PR #767, opened 2026-09-06
Test migrated CSS image and font URLs through HTTP

adamziel:
A full WordPress pull must keep generated stylesheets and serve their image and font URLs from the target site.

The test provisions generated-CSS fixtures in the Elementor and Beaver Builder cache paths, enqueues the first stylesheet through WordPress, then pulls the site and starts its generated PHP runtime. It follows the page to both stylesheets and fetches a real theme font and an SVG through their rewritten URLs. External URLs remain unchanged, and the source stylesheet must not change.

This is a test-only PR based on #764. It uses the supported `localhost` database rewrite target; CSS IP targets have separate downloader coverage in that PR. The full migration passes locally in the Docker E2E environment.

The same end-to-end test fails against the original importer at both stale-CSS-URL assertions.

### DIFF
--- a/tests/e2e/site-registry.json
+++ b/tests/e2e/site-registry.json
@@ -192,6 +192,9 @@
     },
     "empty-geometry-cross-engine": {
       "port": 8141
+    },
+    "generated-css-assets": {
+      "port": 8142
     }
   }
 }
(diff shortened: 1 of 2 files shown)


adamziel:
Moved the end-to-end CSS tests into #764, together with response-cutoff, process-death, mapping-conflict, and preserve-local cases. The feature PR now ships its own E2E coverage, so this separate test-only PR is redundant.

PR #770, opened 2026-09-07
[PHP] Adopt selected multisite tables in a single-site target

adamziel:
Connects selected-site conversion to `pull`, `db-apply` and `apply-runtime`. A selected network site ends as ordinary single-site WordPress, using its existing table prefix.

Create an empty MySQL database named `shop_local`. Replace `alice` with an imported member or content author:

```sh
php reprint.phar pull 'https://network.example/shop/?reprint-api' \
  --secret="$NETWORK_TOKEN" --state-dir=./shop-state --fs-root=./shop-files \
  --target-engine=mysql --target-host=127.0.0.1 \
  --target-user="$MYSQL_USER" --target-pass="$MYSQL_PASSWORD" --target-db=shop_local \
  --new-site-url=http://127.0.0.1:9000 --site-admin=alice \
  --exclude-host-plugins --flatten-to=./shop \
  --runtime=php-builtin --output-dir=./shop-runtime --start-runtime=none

bash ./shop-runtime/start.sh
```

`--site-admin=alice` grants that imported user the administrator role. Site 7 keeps its `_7_` table prefix and `uploads/sites/7` directory. The generated configuration uses the target database credentials, fresh salts, and the imported shared user tables. It does not enable multisite.

The target URL is parsed at the command boundary. IPs, IDNs and trailing-dot hosts are accepted; `127.1` becomes `127.0.0.1`. Text-format limits do not reject a valid destination. Known URL fields use their serializers. Unknown text can remain remote when the fallback cannot safely rewrite it.

An existing target table stops the import before imported SQL runs. Resume keeps the same target database, URL mappings and administrator. Direct `db-pull --sql-output=mysql` is rejected because it skips this target setup.

`db-apply` selects the extra URL parsing from the saved multisite selection. A selected site with no child paths still uses it. Ordinary imports keep the faster STYLE and block-attribute paths from trunk. The mode choice is already on trunk from #796. A SQLite target-apply test also checks it here; the performance report in #795 measures both paths separately.

## Review order

**#770** → #798 → #774 → #775 → #778.

This PR now contains only CLI integration, command validation, runtime configuration, and usage documentation. The URL processors, child-path preflight, and conversion class are already on trunk from #795, #796 and #797. This PR is now based on trunk. The end-to-end scenarios are in the following PR.

## Testing

Open a migrated post, visit site settings and upload a file. Confirm that Network Admin is absent. Try a non-empty target and confirm its tables remain intact. Repeat an interrupted pull with the same options.

Selected-site SQL export still requires a direct PDO MySQL source connection; SQLite child-path lookup in #796 does not add SQLite saved-user locking. Source requirements and remaining limits are in `docs/multisite/USAGE.md`. This work does not remove the existing whole-database table listing or make an arbitrarily large database value fit in memory. If #778 is applied, set a fresh target password as described there before signing in.


packages/reprint-client/src/lib/url-rewrite/class-cautious-url-base-rewrite-mapping.php
```diff
@@ -24,19 +26,75 @@ class CautiousURLBaseRewriteMapping {
...
+            foreach (['http:', 'https:'] as $protocol) {
+                $url = $protocol . substr($origin, strpos($origin, ':') + 1);
+                $parsed = WPURL::parse($url);
+                $parts = parse_url($url);
+                $authority = strtolower($parts['host']) . ( isset($parts['port']) ? ':' . $parts['port'] : '' );
+                $hosts[] = $parsed->host;
+                $hosts[] = $authority;
+            }
+            foreach (array_unique($hosts) as $host) {
+                if ($path_set !== []) {
```
adamziel: Should this be inverted?


```suggestion
            if ($path_set !== []) {
                foreach (array_unique($hosts) as $host) {
```

adamziel: Moved the empty-path check outside the host loop in #794, commit d52d2a11. That PR now contains this code after the split.

PR #778, opened 2026-09-08
[PHP] Exclude shared login credentials from selected-site exports

adamziel:
Keeps shared WordPress password hashes and password-reset keys out of selected-site SQL exports, so a migrated site cannot reuse the source network's login credentials.

Exported users keep their IDs, content links, and site roles. Their `user_pass` becomes the unusable marker `*`, and `user_activation_key` becomes empty. Ordinary single-site exports are unchanged.

## Set the target password

After the full migration, use WP-CLI on the **target**, with the same login passed to `--site-admin`:

```sh
wp --path=/absolute/path/to/target user reset-password alice --skip-email --show-password
```

This generates and displays a fresh target password without sending an email. Keep it private. Other imported users also need new target passwords. Source accounts are unchanged, and Reprint does not accept or save the new login password.

The same password replacements apply to resumed reads, including a tampered cursor requesting an oversized credential field. Exports started with the old rules must start again with a fresh state directory and empty target database; their partial SQL may already contain source credentials.

## Testing

Confirm the source password cannot sign in to the target. Run the command above, then sign in with the new password. The HTTP test also checks that SQL and cursors contain no source password hash or reset key; it fails against #775.

Based on trunk after #798, #774 and #775 merged. Site-admin tokens and tighter shared-file access remain separate work.

Remaining stack: **#778** (based on trunk).


adamziel:
The PHP 8.4 E2E job reached the imported target’s first HTTP request, then the built-in PHP server exited with SIGSEGV. The earlier database and WordPress CLI checks passed. Recorded the crash and exact log in #792; the cause remains unknown. No production workaround or automatic server restart was added.

adamziel:
The same CI run also timed out in source-side setup on PHP 8.5 (before migration), and in the million-site test on PHP 8.1. The PHP 8.1 log took 174.7 seconds to reach the child-path preflight, then the importer reported a 30-second HTTP response timeout; the full test reached its existing 300-second ceiling. The PHP 8.4 failure is the separate SIGSEGV recorded in #792. Retrying these three inspected jobs once without changing code or raising any timeout.

adamziel:
Closing without merging. The PHP 8.4 JIT crash needs to be investigated separately from the selected-site credential change.

PR #783, opened 2026-09-09
[PHP] Retain full progress logs with compact CLI output

adamziel:
Adds `--progress=compact` for short migration tickets while retaining the full JSONL progress stream in `<state-dir>/progress.jsonl`.

Depends on #782. Compact mode prints command starts, stage changes, results, and one final `reprint_report`. Individual skips, grouped warnings, request results, and frequent counters go to the append-only log before ticket filtering. Normal non-TTY output and `--progress=jsonl` still print individual records.

The report groups file warnings by code, with occurrence counts and at most three path examples. Preserve-local skips have a separate count. The summaries survive resume and clear with fresh work or abort. URL-rewrite reports also count tables skipped for lack of a primary key. A successful command can therefore report warnings without changing its exit code.

New logs are private. Failed log writes stop healthy commands rather than silently losing records. The adapter still needs to retrieve this local file if it should be available in MC. `progress.json` remains the live-progress channel; retry scheduling is unchanged.

## Testing

Run a preserve-local file pull with `--progress=compact`. Check that its ticket has no individual skip records, its final report counts them, and `progress.jsonl` contains them all. Repeat with `--progress=jsonl` and check that the individual records remain on stdout. Run compact preflight twice and check that both complete results remain in the log.

After rebasing onto current trunk, the importer suite passes 991 tests with 3 skipped, including warning retention across a failed request and resume, bounded examples, a killed process, and a zero-byte log write. The progress-counter test now also checks compact mode: stdout omits the counters while progress.json and the full log retain the same values. Static analysis and PHP 7.4 compatibility pass. Per-file PHPCS counts do not increase.

### DIFF
--- a/README.md
+++ b/README.md
@@ -937,7 +937,7 @@ Old checkpoints with saved fetch progress lack the completed-byte counts and
 cannot resume in this build. Cleared fetch checkpoints, including those left by
 completed pulls, load with zero completed bytes. Retained indexes are unchanged.
 
-Every command run by `ImportClient` accepts `--progress=auto|tty|jsonl`. The
+Every command run by `ImportClient` accepts `--progress=auto|tty|jsonl|compact`. The
 default `auto` mode uses terminal progress when its output stream is a TTY and
 JSONL otherwise. Use `--progress=tty` to force the terminal presentation when
 output is captured, or `--progress=jsonl` to force structured progress in a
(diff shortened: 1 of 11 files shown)


adamziel:
Replaced by #788, which puts compact output directly on trunk. Reordering the branches caused GitHub to mark this PR as merged into its old report base; compact has not landed on trunk. The stack is now #788 first, then #782 for the optional final report.

PR #795, opened 2026-09-10
Keep child-site URLs remote during selected-site migration

adamziel:
Keeps child-site links at the source when migrating one network site, without sending ordinary imports through the extra parsers.

## Example

Move `https://network.test/shop` to `https://target.test`. `/shop/news` is a separate site:

```html
<!-- Before -->
<a href="https://network.test/shop/article">Shop</a>
<a href="https://network.test/shop/news/article">News</a>

<!-- After -->
<a href="https://target.test/article">Shop</a>
<a href="https://network.test/shop/news/article">News</a>
```

Selected-site imports parse CSS bodies and nested block strings before checking child paths. Path segments and ports must match; `/sites/7` cannot select `/sites/70`. Unknown text on a host with child-path exclusions stays unchanged.

Selected-site block comments use one rewrite walk through their decoded attributes. The same walk handles top-level URL fields and nested strings, then gives the changed tree back to the block parser. The known relative-URL rules and extension hook stay shared with ordinary imports. There is no list of nested attribute paths.

Ordinary imports keep trunk's raw-text paths for STYLE bodies and ordinary block attributes. Encoded block values still use the existing format checks. PR #770 will pass the saved multisite selection to this rewriter, even when the selected site has no child paths. Until then, importer calls keep the ordinary path. No new CLI option or saved mode flag.

JSON script bodies are selected using MIME type and subtype rules, including `text/json` and `model/gltf+json`. Parameters do not affect that choice. Invalid type names stay on the raw-text path. The type check does not copy or parse the parameter tail.

The bounded URL cache is checked before parsing. HTML attributes still escape target quotes at the output boundary. Both are shared with ordinary imports; neither requires the extra CSS or nested-block walk.

## Performance

Selected-site HTML without block comments now uses the HTML parser directly. URL fields share one rewrite function and the existing bounded cache. The cautious text scan runs once on the completed HTML, rather than copying and scanning each token. Documents, incomplete markup and source bases containing literal HTML syntax keep the token-by-token path.

Local PHP 8.5.10 measurements use the same inputs and five checked samples for each build:

| Case | Previous PR head | This change | Trunk |
| --- | ---: | ---: | ---: |
| STYLE bodies, small values | 82 ms | 66 ms | 22 ms |
| 8,192 STYLE elements in one value | 327 ms | 222 ms | 202 ms |
| Nested Divi HTML | 111 ms | 76 ms | 39 ms |
| Repeated nested links | 91 ms | 59 ms | 39 ms |

The large-value case is now in CI. There are 23 checked URL cases, including ordinary imports and selected-site imports. Three cases also supply a child at `/news/` while rewriting `/article/...`; these measure lookup misses. Trunk cannot keep child-site links remote, so its ordinary rewrite is the reference for these identical inputs and expected outputs.

[CI on PHP 8.5.10](https://github.com/WordPress/reprint/actions/runs/34752690258) passes all 23 URL cases. In that run, the large-value case takes 590 ms against trunk's 629 ms. Small STYLE values remain 3.7 times trunk's time; nested HTML is 2.3 times and repeated nested links 1.8 times. Ordinary cases range from 26.8% faster to 0.1% slower than trunk.

The repeated HTML-copy cost is reduced; CSS parsing still costs more than trunk's raw scan on small values. The [full CI comparison](https://github.com/WordPress/reprint/pull/795#issuecomment-5626690274) includes all five samples and peak memory. The local before/after table above isolates the HTML-copy optimization; separate CI runs have different baseline speeds.

The subsequent one-walk block change was also compared locally against `82d3653c` on PHP 8.5.10. All 23 existing benchmark outputs match. A selected-site probe with top-level URLs and nested HTML takes 109 ms instead of 129 ms; with child-path exclusions it takes 111 ms instead of 130 ms. Peak memory stays at 46 MiB. Ordinary benchmark cases stay within 3.1% of the previous head.

## Checks

538 local URL tests, 3,350 assertions, on PHP 8.4 and 8.5. Seven skips and three incomplete tests remain. All 23 benchmark outputs match on the previous head, this change and trunk. PHPStan passes and PHPCS counts do not increase.

Tests cover CSS escapes, child paths, ports, relative-URL cache separation, unchanged attribute quoting, incomplete HTML, 5,000 blocks and 3,000 STYLE elements under a 64 MiB limit. No new cache. The source-base HTML delimiter check uses a byte count instead of copying the matching suffix. Rewriting still holds one complete database value. Cache budgets remain 4 MiB for URLs and 32 MiB for values.

Stack: **#795** → #796 → #797 → #770 → #798 → #774 → #775 → #778.

### DIFF
--- a/packages/reprint-client/src/lib/url-rewrite/class-structured-block-markup-url-processor.php
+++ b/packages/reprint-client/src/lib/url-rewrite/class-structured-block-markup-url-processor.php
@@ -16,15 +16,11 @@ use WordPress\DataLiberation\URL\WPURL;
  * [vc_video link="https:\/\/source.example\/media\/video.mp4"]
  * ```
  *
- * Declared URL fields are final after parsing, including rejected values.
- * The fallback never scans them again. For example, with A -> B and B -> C
- * mappings, a parsed URL from A must stop at B.
- *
- * The fallback changes only source-base bytes inside unparsed values. HTML
- * attribute values and raw-text element bodies use the HTML decoder/setter,
- * so changed values can have different quotes, entities, or newline spelling.
- * Ordinary text and comments keep raw-token replacement: shortcode syntax
- * such as the example above never passes through the HTML text encoder.
+ * Decoding and serializing that URL could change its slashes, quotes,
+ * entities, query, or fragment. After structured URLs have been handled,
+ * replace_url_bases_in_current_token() changes only configured source-base
+ * bytes in the raw token. The surrounding bytes never pass through the HTML
+ * text encoder.
  *
  * @TODO Contribute this structured processor back to the PHP toolkit after
  *       its Reprint behavior has stabilized.
@@ -51,19 +47,8 @@ class StructuredBlockMarkupUrlProcessor extends BlockMarkupProcessor {
 	private $css_url_processor;
 	private $css_url_processor_updated;
 
-	/** @var CSSURLProcessor|null CSS in the current STYLE body, separate from its style attribute. */
-	private $style_element_processor;
-	/** @var bool Whether a CSS URL edit must be written back to the current STYLE body. */
-	private $style_element_updated = false;
-	/**
-	 * Top-level block fields reserved for URL parsing in the current token.
-	 * For wp:image's `url`, a parse failure still excludes it from the fallback.
-	 * An unknown field is reserved only if its whole value parsed as a URL.
-	 * Nested fields stay with the caller's format inference.
-	 *
-	 * @var array<string, true>
-	 */
-	private $url_block_attribute_keys = array();
+	/** @var bool Whether the CSS parser reads a STYLE body instead of a style attribute. */
+	private $in_style_element = false;
 
 	/**
 	 * The list of names of URL-related HTML attributes that may be available on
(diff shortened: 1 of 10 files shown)


adamziel:
The new benchmark in #799 measures the fast path removed here. I ran the same harness against this PR's `3f0b1a94` PHAR and trunk's `f8cdc5e4` PHAR from [the CI build](https://github.com/WordPress/reprint/actions/runs/34585100193).

Local native PHP 8.4.5 results, using the median of five samples for 128 values with 32 entries each:

| Case | Trunk | #795 | Time ratio |
| --- | ---: | ---: | ---: |
| Nested Divi HTML, distinct URLs | 61 ms | 928 ms | 15.1× |
| Nested Divi HTML, repeated URLs | 60 ms | 454 ms | 7.5× |
| Plain URLs in nested block attributes | 60 ms | 82 ms | 1.4× |
| STYLE bodies | 38 ms | 987 ms | 26.1× |
| Nested HTML with escaped quotes | 1,067 ms | 936 ms | 0.9× |

Every output passed its content check on both builds. The first run, in the opposite build order, also showed about 15× for nested HTML and 7.6× for repeated URLs. These are URL rewrite timings, not whole-import timings or PHP.wasm results.

The ordinary nested-HTML cases take the old raw-JSON fast path. The escaped-quote control already used parsing. Removing the fast path has a substantial cost for ordinary Divi content; the existing memory test did not measure that cost. #799 adds these rows to the per-PR performance report without changing this PR's rewriter.

PR #802, opened 2026-09-11
Pull Windows-hosted WordPress sites onto Linux

adamziel:
Pull Windows-hosted WordPress sites onto Linux using PHP file functions, with no FFI requirement.

Path conversion takes both the path and `path_format` (`unix` or `windows`). The source reports its format in preflight; the client saves it for later commands and resume. For `D:\photos`, Unix mode keeps the backslash and Windows mode returns `D:/photos`. Local remap targets use the local format. Comparisons no longer change separators. Link targets also use the source link's directory. Older servers retain the Unix-only contract; invalid format values fail. Inline comments explain these rules with concrete examples.

Drive paths keep their drive letter under the local file root. Shares map under `UNC/SERVER/SHARE`. The Windows source resolves ordinary drive/share namespace prefixes, relative selections and filename case before the client builds path rules. It retains selected links so `--no-follow-symlinks` can inspect them. Missing tails remain valid for exclusions and previously pulled paths.

The FFI filesystem class and stream wrapper are removed. Long UNC paths use a PHP-supported namespace spelling only when ordinary metadata lookup fails; that spelling stays out of indexes and cursors. PHP still enforces `open_basedir`. Junctions remain links even though PHP leaves their `lstat()` type bits at zero. Root-relative link targets are resolved through PHP's link target data. Windows CI runs with the FFI extension absent.

Unreadable names stop the pull. PHP can return an ordinary sibling's metadata for a name ending in a dot or space, so those names are rejected before I/O, including when found inside a selected parent. Rename them before starting a fresh pull. Volume GUID and `GLOBALROOT` inputs require a drive-letter or UNC spelling. A drive-relative input for another drive requires a full path. Links whose targets PHP cannot read or resolve stop with instructions to recreate them. NTFS alternate data streams are not migrated, and Linux filename limits produce errors rather than renamed or omitted files.

A read error must remain an error after resume. The index keeps an unreadable entry at its cursor and retains a selected link until inspection succeeds. The client stops on file-fetch exceptions instead of repeating the same failed request. The existing Windows creation-time limitation is documented beside the PHP metadata reads.

## Testing

All 32 CI checks pass at `7cdbed37`. The [Windows workflow](https://github.com/WordPress/reprint/actions/runs/35084545789) passes complete WordPress migrations from a drive and a network share into Linux. It checks hashes, empty directories, database rows, URL rewriting, saved source format, and raw and flattened runtimes. Separate pulls cover punctuation, Unicode, long drive and UNC paths, drive/share aliases, relative inputs, case-sensitive names, junctions, link remaps, no-follow behavior, and repeated failures at the documented limits. Reader checks also cover chunk resume, error-cursor resume, overlapping roots and `open_basedir`.

The [PHP-only probe](https://github.com/WordPress/reprint/actions/runs/35080086796) recorded the filename and link limits with FFI absent. The new file-fetch exception test failed before the client change and passes after it; the real Windows migration also exercises that failure over HTTP. The local filesystem suite runs 558 tests with one platform-specific skip. PHPStan and PHP compatibility checks pass, and changed files do not increase PHPCS counts.


adamziel:
The replacement stack is open, in merge order:

#806 → #807 → #808 → #809 → #810 → #811

Each PR has its own behavior, tests, and inline comments. #811 contains only the final migration workflow, fixtures, and user documentation. The combined code retains this PR's migration cases, uses the current POST request API, and adds a real HTTP check for Windows selection resolution.

CI is still running on the upper PRs. This PR remains open as the reference until those checks pass; use the smaller stack for review.

adamziel:
Replaced by six smaller PRs, in merge order: #806 → #807 → #808 → #809 → #810 → #811. All six are ready for review and all their CI checks pass, including both complete Windows-to-Linux migrations. Tests and inline comments stay with each behavior. This branch is retained as the original reference. Nothing has been merged.

PR #805, opened 2026-09-15
Require explicit permission for HTTP connections

bcotrim:
Reject HTTP remote Reprint API URLs unless `--allow-unsafe-http` is passed.
Keep `--force-http` as an alias. 
Optional `ImportClient` constructor settings use the existing options-array pattern, with `allow_http` internally; known options are validated and unrelated keys are ignored.

## Testing instructions

Use a local Studio site with Reprint Server. Confirm HTTPS preflight succeeds with and without the flag, HTTP without permission fails before creating state, and both HTTP flags work when Studio HTTPS is disabled. With Studio HTTPS enabled, an explicitly allowed HTTP request must report its redirect rather than follow it. Run fresh and repeat full pulls over HTTPS and explicitly allowed HTTP, then omit the flag while reusing HTTP state and confirm rejection.


packages/reprint-client/src/import.php
```diff
@@ -619,6 +647,15 @@ public function __construct(
         $this->state = new PullState();
     }
 
+    public static function validate_remote_reprint_api_url_transport(string $remote_reprint_api_url, bool $allow_http): void
+    {
+        if (!$allow_http && strncasecmp($remote_reprint_api_url, 'http://', 7) === 0) {
+            throw new InvalidArgumentException(
+                'HTTP is insecure. Use HTTPS or pass --allow-unsafe-http.'
```
adamziel: Let's make this error message more specific. The user who sees it should understand what they did wrong, why, and what's the next step. It has a little bit of that already since it explains the CLI switch to use. What would make it even more helpful is something like "The remote site URL you've provided uses a HTTP protocol. HTTP is unencrypted and exporting the entire site over it would expose all its data, including passwords, to eavesdropping. Provide a HTTPS URL to keep your data secure, or pass the `--allow-unsafe-http` option to accept this risk.".

adamziel:
Thank you @bcotrim!

PR #809, opened 2026-09-16
Resolve Windows source paths once and reuse them across pulls

adamziel:
Saves Windows source path resolutions in pull state so later pulls with the same expanded input list need no path-resolution request.

## Before and after

Two remap sources, two includes, and two exclusions previously caused six requests on every pull. The first pull now resolves them in one POST request and saves the input/result pairs in `state.json`. Later pulls load that file and reuse the results. Local destination paths stay on the client.

The latest successful batch is reused while the source inputs match, after WordPress tokens have been expanded. A changed input list or token value sends one new batch. A new preflight timestamp or a changed local destination does not cause another request. The cache survives command resets. Existing remap and selection fingerprint checks still apply.

`D:photos` keeps the absolute path selected by its first resolution even if the source process later changes its working directory. This is deliberate: an unchanged selection continues to name the saved source path. A new state directory starts without cached results.

## Request and state format

The request carries a JSON list of base64 paths in one form field. PHP's `max_input_vars` therefore cannot cut off a long list of selections. Results stay in request order. The cache stores both lists as base64 so JSON preserves filename bytes.

One rejected path fails the whole batch. Failed resolution leaves the previous saved batch intact. The client does not apply partial results or retry paths one at a time. Empty selections, Unix sources, and older sources without this endpoint send no path-resolution request. Older pull state loads with an empty cache.

## Source path rules

The source host supplies relative-path context. The endpoint handles ordinary drive and UNC spellings, supported namespace prefixes, relative paths, and filename case. It retains missing tails for exclusions and deleted paths. It stops resolving at a link so a no-follow selection still sees a junction. Device inputs, unsupported volume namespaces, and names PHP cannot read exactly fail with an instruction to change the selection or source name.

Preflight advertises the endpoint only where it is available. Path resolution remains in pull; no new options are passed to preflight. The existing plugin authentication protects the endpoint, and restricted multisite requests cannot use it.

## Testing

Windows CI runs the real client against the real PHP HTTP endpoint and counts requests. The batching tests cover combined options, duplicates, the `--only` alias, both file filters, empty selections, older sources, Unix backslashes, and whole-batch rejection. A 1,205-path selection runs with `max_input_vars=1000`. The endpoint tests retain all 22 Windows path cases and add malformed batch inputs.

The cache tests launch a separate client process for each path-preparation run. They cover reuse from `state.json`, fresh preflight data, command resets, changed local targets, changed source inputs and tokens, rejected replacement batches, and a changed source working directory. The regression failed against the previous implementation with “expected 0 resolver requests after loading pull state; got 1.” State tests also cover older files and byte-preserving JSON round trips.

Local validation passes 315 focused tests with one platform skip, PHPStan, and PHP compatibility checks. Per-file production PHPCS counts are unchanged.

## Stack

Part 4 of the split of #802. #808 has merged; this PR targets `trunk`. Full Windows-to-Linux WordPress migration tests follow in #811.

Remaining stack: #809 → #810 → #811.


adamziel:
Closing without merging. Windows selections will use the existing Unix-style input contract: full absolute paths or WordPress path tokens. The separate path-resolution request and persistent cache are dropped. #810 now targets trunk with only its runtime-mapping changes; #811 will verify that smaller stack. This branch is retained for reference.

PR #810, opened 2026-09-16
Use pull path mapping throughout WordPress runtime setup

adamziel:
WordPress runtime setup uses the same remote-to-local path mapping as the file pull.

## Before and after

A source file at `\\server\share\Sites\index.php` is downloaded under `--fs-root/UNC/SERVER/SHARE/Sites/index.php`. Runtime setup previously built its own path and could look somewhere else. It now uses the download mapper, so both stages select the same local file.

The same mapping applies to the document root, WordPress core, content and plugin directories, uploads, a parent wp-config.php, and the default SQLite location. Explicit remaps stay consistent across download, content merge, flattening, and runtime setup.

## Testing

The focused tests cover Unix roots, Windows drive paths, and UNC paths for the raw runtime and parent wp-config.php layout. After removing the dependency on #809, 151 focused tests pass locally with 637 assertions. PHPStan passes; production PHPCS counts are unchanged.

## Stack

Targets `trunk`, following the merged #806, #807, and #808. #809 is dropped: this PR contains no path-resolution endpoint or cache. Source selections use the existing absolute-path and WordPress-token rules. #811 tests complete Windows-to-Linux migrations on top of this PR.

Remaining stack: #810 → #811.


adamziel:
The initial Playground shard 1/4 run reported the WP Engine fixture as `other` before runtime setup. The final stack passed the same suite with identical production code. An unchanged rerun of the failed jobs also passed: https://github.com/WordPress/reprint/actions/runs/35095049544/attempts/2. No test or production code was changed for that rerun.

PR #832, opened 2026-09-22
Authenticate with public keys on hosts that have OpenSSL

fredrikekelund:
> [!NOTE]  
> This mega PR contains **all** the changes needed to make public-key cryptography the main authentication mechanism for Reprint, across client, server, and WP plugin. It is intended for cursory review, after which I can break it down into a stack of PRs that's easier to review.

Adds public-key authentication to the Reprint API so the client no longer needs a shared token that must be copied out of the site's settings page. The connection-token (HMAC) scheme stays; which one a site uses is decided by its host.

## How it works

**The host's PHP picks the scheme; nothing configures it.** `Utils::key_auth_required()` is `function_exists('openssl_verify')`. A host that has it accepts only key-signed requests; a host without it accepts only connection tokens, exactly as today. The two schemes never coexist on one site, and the client never retries with the other one.

**Key signatures reuse the HMAC envelope.** `PublicKeyClient`/`PublicKeyServer` sign and verify the same message HMAC already covers (method, URL, nonce, timestamp, cursor, content hash), carried in `X-Auth-*` headers instead of the HMAC ones. The only new material is the signature itself (RSA-2048, SHA-256). `EnvelopeSigner` is the interface both `Site_Export_HMAC_Client` and `PublicKeyClient` implement, so the push stream client accepts either through its `envelope_signer` option (formerly `hmac_client`).

**`RequestAuthenticator` is the single server entry point.** It holds the stored token and the enrolled public keys and dispatches by the host rule. Refusals carry a reason code — `requires_key_auth`, `requires_token_auth`, `unknown_key`, `auth_failed` (403) and `not_configured` (503, a key host with no keys enrolled) — which the client maps to a message saying what to run. `HMACServer` applies the same host rule on its own, so a direct caller on an OpenSSL host is refused rather than silently accepting a token; **embedders that call `HMACServer` directly (Jetpack does) should adopt `RequestAuthenticator` and key storage when they bump the package.**

**Enrollment lives on the settings page.** Tools → Reprint Server (network settings on multisite) gets a paste-a-key form and a key table with per-key push grant and removal. Keys are stored in the `reprint_server_public_keys` option, or in a `public-keys.php` file beside the plugin when one exists (the file wins, mirroring `secret.php`). On a key host the stored token is shown as not accepted.

**The client finds its own key.** Credential resolution per command is `--secret` → `--private-key=PATH` → `<state-dir>/remotes/<md5-of-url>/key.pem`. `reprint keygen <url> --state-dir=DIR` generates a key beside the site's state and prints the public half; `reprint pull` with no credential does the same and stops with exit code 4 so the user can enroll it. Deleting the state directory destroys the private half, so the enrolled public key stops working; the settings page's Remove button revokes it explicitly.

## Also in this PR

- The pull cursor was sent twice, as the `X-Export-Cursor` header and in the request body, and the server reads the body copy. Key signatures cover the cursor, so a proxy that strips the header left a valid request that failed verification. The header is gone; the body copy is already covered by the signed content hash.
- `PublicKeyClient::generate_keypair()` works on PHP builds without a default `openssl.cnf` (setup-php, most Windows PHP), which `openssl_pkey_new()` otherwise refuses.
- The e2e suite authenticates with keys on the ordinary FPM pools and adds an `e2e-hmac-only` pool (`disable_functions=openssl_verify`) so both host kinds run end to end; CI harness scripts, the benchmark and the Windows migration jobs sign with an enrolled key.
- The client package requires `ext-openssl`; the server package suggests it. `composer.lock` records both.

## Upgrade note

Installing this plugin version on a host with OpenSSL disables token authentication there: every request gets `not_configured` until a key is enrolled. Nothing is deleted, and enrolling a key restores access at once. Update client and plugin together.


packages/reprint-client/src/import.php
```diff
@@ -186,8 +186,20 @@ class ImportClient
         "flat-docroot",
         "merge-wp-content",
         "apply-runtime",
+        "keygen",
     ];
 
+    /** Commands that authenticate to the remote site and therefore need a credential. */
+    public const REMOTE_COMMANDS = [
```
adamziel, line 193: Don't we have a list like that somewhere? How do we know when to use HMAC?

packages/reprint-client/src/import.php
```diff
@@ -2430,6 +2494,205 @@ public static function remote_state_directory_path(
...
+            throw new RuntimeException(
+                "A key already exists at {$path}. It may still be enrolled and in use. Pass --force to replace it."
+            );
+        }
+        $directory = dirname($path);
+        if (!is_dir($directory) && !mkdir($directory, 0700, true) && !is_dir($directory)) {
+            throw new RuntimeException("Could not create {$directory}.");
+        }
+        [$private_key_pem, $public_key] = \WordPress\Reprint\Server\PublicKeyClient::generate_keypair();
+        // Write a fresh sibling file and rename it over the target. Rewriting
```
adamziel, line 2526: afair there's a helper for the `create .tmp -> rename` part of it

